PR #165, opened 2015-01-09
Add MSDN-SETCAPTURE.

scheib:


tobie:
Thanks for the pull request. I'm wondering how pertinent it is to include it in the database rather than just include it in a local bibliography (easy depending on what tools you use).

Thoughts? /cc @darobin, @marcoscaceres

scheib:
Thanks. I'm using ReSpec for Pointer Lock spec. The MSDN reference is the source of the setCapture feature implemented in multiple browsers and that there is no spec for it, so it seems reasonable that more than PointerLock may eventually reference it. But, I think a local biblio reference would be fine.

The ReSpec doc sort of pointed me towards adding the reference to specref
http://www.w3.org/respec/guide.html#references
I'm not aware of how to make local biblio references in ReSpec. Perhaps that should be a feature, or documented if already one.

tobie:
Given there's already [MDN-SETCAPTURE](http://www.specref.org/?q=MDN-SETCAPTURE), we might as well have MSDN-SETCAPTURE.

tobie:
@scheib for future ref, local biblios are easy to build in ReSpec, just use the [`localBiblio` property](http://www.w3.org/respec/ref.html#localbiblio). Great for obscure references no one else cares about or when there's a deadline and a PR doesn't get pulled in fast enough.

PR #166, opened 2015-02-02
add entry for WAKE-LOCK-USECASES

dontcallmedom:
for use in https://github.com/w3c/wake-lock/


tobie:
@dontcallmedom the spec you're pointing to references `WAKE-LOCK-USECASES` while the spec's shortname is `WAKE-LOCK-USE-CASES`. So you need to fix the reference in that spec or provide a alias here in the form of:

```js
{
    "WAKE-LOCK-USECASES": { "aliasOf": "WAKE-LOCK-USE-CASES" },
  // …
}
```

dontcallmedom:
I see; I'll get the spec fixed, sorry for the noise :)

PR #179, opened 2015-04-21
Fix security issue

azu:
Hi! Thanks for useful database.

I found a XSS in http://www.specref.org/ .

specref 2015-04-21 14-00-08

http://www.specref.org/?q=%3Cscript%3Ealert%281%29%3C%2Fscript%3E%3Cs%3E

This pull request fixes XSS.

- use `$.text` instead of `$.html`

### DIFF
--- a/css/search.css
+++ b/css/search.css
@@ -70,6 +70,7 @@ strong.highlight {
     margin-top: 10px;
     color: #777;
     font-size: smaller;
+    white-space: pre-wrap;
 }
 
 h1 a {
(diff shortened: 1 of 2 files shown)


css/search.css
```diff
@@ -70,6 +70,7 @@ strong.highlight {
     margin-top: 10px;
     color: #777;
     font-size: smaller;
+    white-space: pre-wrap;
```
azu: replacement for `<br>`

PR #183, opened 2015-05-01
Moar HTTPS and use /multipage/ for HTML

annevk:


tobie:
Happy to merge this but heads-up that some of these will be overwritten by the auto-update scripts.

tobie:
@annevk, here's what the auto-update script does: 97a437d, the simplest way to fix this is in the script itself, imho, by adding them one by one to the [ED_DRAFTS dictionary](https://github.com/tobie/specref/blob/master/scripts/rdf.js#L41-L43).

domenic:
Can we make the script just auto-rewrite all github.io URLs to https?

tobie:
Yup, PR on its way.

tobie:
Fixed in #185

PR #191, opened 2015-06-18
ECMA 262 6th edition is a standard

plehegar:
We should stop linking to 5.1 as far as I know.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -596,11 +596,7 @@
         "aliasOf": "RFC6090"
     },
     "ECMA-262": {
-        "title": "ECMA-262 6th Edition, The ECMAScript 2015 Language Specification",
-        "date": "June 2015",
-        "href": "http://www.ecma-international.org/ecma-262/6.0/index.html",
-        "status": "Standard",
-        "publisher": "ECMA"
+        "aliasOf": "ECMASCRIPT"
     },
     "ECMA-262-6.0": {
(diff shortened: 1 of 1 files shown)


refs/biblio.json
```diff
@@ -596,17 +596,33 @@
...
-    "ECMA-262-51": {
+    "ECMA-262-2015": {
         "aliasOf": "ECMA-262"
     },
+    "ECMA-262-5.1": {
+        "date": "June 2011",
+        "href": "http://www.ecma-international.org/publications/standards/Ecma-262.htm",
+        "publisher": "ECMA",
+        "title": "ECMAScript Language Specification, Edition 5.1",
+        "deliveredBy": [
```
tobie: You might want to remove the "deliveredBy" props, which are really here to identify WG (not publishers).

refs/biblio.json
```diff
@@ -596,17 +596,33 @@
         "aliasOf": "RFC6090"
     },
     "ECMA-262": {
-        "date": "June 2011",
-        "href": "http://www.ecma-international.org/publications/standards/Ecma-262.htm",
+        "title": "ECMA-262 6th Edition, The ECMAScript 2015 Language Specification",
```
tobie: I would suggesting aliasing the whole of `[ECMA-262]` to `[ECMASCRIPT]` and adding "https://people.mozilla.org/~jorendorff/es6-draft.html" as an `"edDraft"`

PR #219, opened 2015-09-03
Add Web Bluetooth to the specref database.

jyasskin:


tobie:
Should be deployed within minutes.

PR #221, opened 2015-10-16
Update ES262 and ES402

domenic:


domenic:
Would you like to maintain an alias from ECMAS-402 to ECMA-402? It seems like a clear typo.

tobie:
Mmm. I don't have an effective means of checking whether someone still relies on it. So yeah, I'd rather we left it around, as an alias, preferably.

domenic:
OK cool, pushed with an alias.

tobie:
Thanks!

PR #226, opened 2015-11-03
Add Dave Longley to list of spec authors for JSON-LD.

msporny:
The JSON-LD specs in the specref database don't list Dave Longley because the entries are created by just looking at the Editor fields. In the particular case of the JSON-LD specs, Dave Longley contributed a non-trivial number of fundamental concepts and algorithms to the core of the language and I had realized a while ago that he was erroneously not listed at all in the spec refs. Finally getting around to fixing this issue.

### DIFF
--- a/refs/w3c.json
+++ b/refs/w3c.json
@@ -27116,6 +27116,7 @@
     "json-ld": {
         "authors": [
             "Manu Sporny",
+            "Dave Longley",
             "Gregg Kellogg",
             "Markus Lanthaler"
         ],
@@ -27138,6 +27139,7 @@
             "20130411": {
                 "authors": [
                     "Manu Sporny",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Markus Lanthaler"
                 ],
@@ -27154,6 +27156,7 @@
             "20130910": {
                 "authors": [
                     "Manu Sporny",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Markus Lanthaler"
                 ],
@@ -27170,6 +27173,7 @@
             "20131105": {
                 "authors": [
                     "Manu Sporny",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Markus Lanthaler"
                 ],
@@ -27186,6 +27190,7 @@
             "20140116": {
                 "authors": [
                     "Manu Sporny",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Markus Lanthaler"
                 ],
@@ -27204,6 +27209,7 @@
     "json-ld-api": {
         "authors": [
             "Markus Lanthaler",
+            "Dave Longley",
             "Gregg Kellogg",
             "Manu Sporny"
         ],
@@ -27232,6 +27238,7 @@
             "20130516": {
                 "authors": [
                     "Markus Lanthaler",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Manu Sporny"
                 ],
@@ -27248,6 +27255,7 @@
             "20130910": {
                 "authors": [
                     "Markus Lanthaler",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Manu Sporny"
                 ],
@@ -27264,6 +27272,7 @@
             "20131105": {
                 "authors": [
                     "Markus Lanthaler",
+                    "Dave Longley",
                     "Gregg Kellogg",
                     "Manu Sporny"
                 ],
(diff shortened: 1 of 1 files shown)


msporny:
Hey @tobie - will these entries in the w3c.json spec ref database be overwritten when the automatic update process runs? If so, is the right thing to do to update refs/biblio.json?

tobie:
Yeah. They'll be overwritten. The right thing to do is to get it fixed upstream, as fixing this in specref is going to break in all sorts of unpredictable ways.

tobie:
Note I know this is not optimum but I don't have the resources at present to make this better.

msporny:
So, the specs are at TR, so I can't get it "fixed" upstream, can I? Could I copy/paste all corrected JSON-LD and JSON-LD-API entries to refs/biblio.json instead?

If not, what would need to change in the specref processor to support overriding an authors list? I might write that patch if I knew where to do the patch.

tobie:
> So, the specs are at TR, so I can't get it "fixed"
> upstream, can I?

Mmm. That's a good question. Probably too painful for just an errata.

> Could I copy/paste all corrected JSON-LD and
> JSON-LD-API entries to refs/biblio.json instead?

Possibly, but that might break at the next update in ways I can't predict (but which could be tested). 

> If not, what would need to change in the specref > processor to support overriding an authors list?

Plenty of things in rtf.js. Not sure which ones. You'll need to test to see what happens when the update scripts are run. 

> I might write that patch if I knew where to do the > patch.

I'd happily fix it myself if that file wasn't a pile of unpredictable spaghetti code of my own doing.

msporny:
Alright, then here's a proposal:

1. Specref (specifically, scripts/rdf.js) supports a file called "refs/w3c-overrides.json". This file contains a structure very similar to w3c.json, except it only contains values that are meant to be overridden in the final w3c.json output file.
2. After all the W3C specs have been pulled from the W3C website and rdf.js has finished processing, the final phase is to apply refs/w3c-overrides.json. This enables specref to override anything that needs to be overridden on a W3C spec.

The upside here is that we don't need to "fix" W3C's TR spec publication tool and then figure out how to merge authors/editors into a coherent list.

If you are ok w/ this proposal, I can spend some time writing it and sending a pull request for it.

tobie:
That sounds like a plan. You might wan't to generalize it by adfing it to the runner.

darobin:
On 03/11/2015 00:47 , Manu Sporny wrote:
> So, the specs are at TR, so I can't get it "fixed" upstream, can I?

If there is a genuine error in TR, and it doesn't affect conformance, I
would at least as a first step try to appeal to the PTB to get it fixed
in place (or even issue a 2E).

-- 
• Robin Berjon - http://berjon.com/ - @robinberjon
• http://science.ai/ — intelligent science publishing
•

msporny:
@darobin it's not an error per-sec. The W3C TR spec RDF that's published doesn't include 'author' information for specs (it only includes editors). Since that information isn't included, we can't add a feature in specref to generate the references by merging the editors and the authors. Even if we did that, we may need to override the editor/author information anyway. I'll try to submit a bug on the TR spec RDF that's generated later this week, but don't expect it to be fixed with the new move toward the W3C HTTP API. Perhaps the proper fix is to upgrade specref to use the new W3C HTTP API (which does include author information, IIRC). Thoughts @vivienlacourba?

darobin:
On 03/11/2015 10:22 , Manu Sporny wrote:
> @darobin <https://github.com/darobin> it's not an error per-sec. The W3C
> TR spec RDF that's published doesn't include 'author' information for
> specs (it only includes editors). Since that information isn't included,
> we can't add a feature in specref to generate the references by merging
> the editors and the authors. Even if we did that, we may need to
> override the editor/author information anyway. I'll try to submit a bug
> on the TR spec RDF that's generated later this week, but don't expect it
> to be fixed with the new move toward the W3C HTTP API. Perhaps the
> proper fix is to upgrade specref to use the new W3C HTTP API (which does
> include author information, IIRC). Thoughts @vivienlacourba
> <https://github.com/vivienlacourba>?

The TR RDF is certainly horrible, instead of codifying exceptions you
would certainly be better off just using the JSON API.

https://github.com/w3c/node-w3capi is your friend.

-- 
• Robin Berjon - http://berjon.com/ - @robinberjon
• http://science.ai/ — intelligent science publishing
•

tobie:
I welcome a pull request on that topic. But be warned there are plenty of edge cases.

tobie:
Closing. PR welcomed when you have the time.

PR #234, opened 2016-02-18
Add WEB-BACKGROUND-SYNC

nicolassatragno:
I'd like to add the draft for background sync to the database. What do you think?


tobie:
Deployed: http://www.specref.org/?q=WEB-BACKGROUND-SYNC

PR #248, opened 2016-03-04
Fix authors and titles of old DOM specs.

tobie:
/cc @annevk


annevk:
Heh, couldn't resist huh? 😊

tobie:
Well, it is one of the rare cases where fixed references like that are _actually_ useful. kind of a shame if it doesn't work.

PR #249, opened 2016-03-12
Add the promises-guide.

jyasskin:
@domenic


tobie:
We should pull-in the findings automatically.

Issue #184, opened 2015-05-01
[[FULLSCREEN]] should map to fullscreen.spec.whatwg.org


annevk:
No other specification is maintained. We could call the other version LEGACY-FULLSCREEN or W3C-FULLSCREEN? Happy to write a pull request for this.

tobie:
So this implies also handling this somehow in the auto-update scripts or it'll get overwritten when the next auto-update kicks in.

annevk:
Can you give me a pointer? Happy to do that.

tobie:
The auto-update scripts are kind of a giant ball of Scotch tape at this point, so I'm not 100% sure what the solution is. Not to mention that W3C has two completely different ways of indicating retired specs, one which I just found out about now. FYI, the file that handles W3C updates is "/scripts/rdf.js".

Simply renaming things doesn't work as it's immediately overwritten by the auto-update script.

Ideally, W3C would publish the specs that are superseding retired specs, but I'm not holding my breath.

I need to think about this a bit more. Would <3 input from @darobin's and @marcoscaceres on this topic.

tobie:
Fixed the isRetired issue in #186. Intend to use this as a way to prevent overwriting.

marcoscaceres:
hmmm... I guess the best thing to do right now is to have a "whatwg-fullscreen" ref. There is no much we can do about W3C squatting "fullscreen" :( I think that will be fine, as the fullscreen spec has been gutted and points anyone who lands there to the right place.

tabatkins:
@marcoscaceres Nah, that's just a short-term solution. As tobie alluded to, when we know that a spec is retired, we can avoid having the auto-updater overwrite a same-name spec from a different source.

marcoscaceres:
Ok, we should work with the w3c team then.

tobie:
Well, we have this data; a bunch of specs--including this one--are already marked as retired. Now we just have to account for this information in order to avoid overwriting references that have been manually modified.

tabatkins:
Looks like it's just a matter of, at `rdf.js:142`, checking if the rdf ref is retired, and the `current` ref is `!isGeneratedByThisScript()`, and only doing the work in that conditional if both of those are false.

annevk:
[...]

PR #250, opened 2016-03-15
Fix stale refs (DOM only)

tobie:
OK, this is an attempt at manually fixing the problem of stale W3C references now edited within WHATWG. See #184 for discussion.

This PR focuses only on DOM.

/cc @tabatkins @domenic @annevk


tobie:
OK, so this is breaking stuff in prod in weird ways.

tobie:
Had to backtrack out of it.

PR #261, opened 2016-03-22
Expose isbn and pages props.

tobie:
/cc @arronei

### DIFF
--- a/lib/bibref.js
+++ b/lib/bibref.js
@@ -13,6 +13,8 @@ var PROPS =  [
     "date",
     "status",
     "publisher",
+    "pages",
+    "isbn",
     "aliasOf",
     "obsoletes",
     "obsoletedBy",


tobie:
Curious as to whether there's interest from API consumers to use that info for book references. /cc @tabatkins, @marcoscaceres, @halindrome

marcoscaceres:
I've personally never found myself needing to cite a book or journal in a spec. But it could be useful.

tabatkins:
I'm happy to output as specific/useful a biblio reference as possible, so if there's more information about ISBN and pages, I can def use that.

tobie:
http://www.specref.org/?q=0-13-000027-2

marcoscaceres:
@tabatkins, let me know where you putting page numbers and ISBN (I.e., what bibliographical style you are using) and I'll do the same on ReSpec side.

Issue #147, opened 2014-10-21
Convert legacy HTML references to structured data.


tobie:
These references are still in need of conversion:
- [x] [ABC](http://specref.jit.su/bibrefs?refs=ABC)
- [x] [AD-INDUSTRY](http://specref.jit.su/bibrefs?refs=AD-INDUSTRY)
- [x] [AGBDL](http://specref.jit.su/bibrefs?refs=AGBDL)
- [x] [ANSI-X9-44-2007](http://specref.jit.su/bibrefs?refs=ANSI-X9-44-2007)
- [x] [ARABIC-TYPO](http://specref.jit.su/bibrefs?refs=ARABIC-TYPO)
- [x] [BBC-WP193](http://specref.jit.su/bibrefs?refs=BBC-WP193)
- [x] [BBC-WP194](http://specref.jit.su/bibrefs?refs=BBC-WP194)
- [x] [BONDI-ARCH-SECURITY-11](http://specref.jit.su/bibrefs?refs=BONDI-ARCH-SECURITY-11)
- [x] [BONDI-FEATURES1-11](http://specref.jit.su/bibrefs?refs=BONDI-FEATURES1-11)
- [x] [CC-ABOUT](http://specref.jit.su/bibrefs?refs=CC-ABOUT)
- [x] [CC-CHOOSE](http://specref.jit.su/bibrefs?refs=CC-CHOOSE)
- [x] [CEA-2014](http://specref.jit.su/bibrefs?refs=CEA-2014)
- [x] [CHARSETS](http://specref.jit.su/bibrefs?refs=CHARSETS)
- [x] [COAP](http://specref.jit.su/bibrefs?refs=COAP)
- [x] [COLORIMETRY](http://specref.jit.su/bibrefs?refs=COLORIMETRY)
- [x] [CONSENT-EU-WP187](http://specref.jit.su/bibrefs?refs=CONSENT-EU-WP187)
- [x] [CVE-2009-0217](http://specref.jit.su/bibrefs?refs=CVE-2009-0217)
- [x] [Coloring-RDF](http://specref.jit.su/bibrefs?refs=Coloring-RDF)
- [x] [Context-SW](http://specref.jit.su/bibrefs?refs=Context-SW)
- [x] [DC-TERMS](http://specref.jit.su/bibrefs?refs=DC-TERMS)
- [x] [DES](http://specref.jit.su/bibrefs?refs=DES)
[...]

PR #262, opened 2016-03-22
Convert all "B" items of legacy strings to structured data

arronei:
* convert all the "B" items for issue #147 

@tobie Take a look at BONDI-* items I'm not sure if those are exactly right with the version stuff.


tobie:
I don't think there's a benefit in using `"versions"` for this legacy stuff. Makes it way harder to review, too.

arronei:
Cool I will simplify this down then. It will also fix my build error.

tobie:
These two things are connected.

PR #265, opened 2016-03-27
Tightened up the DCMI terms in refs.json

iherman:
DCMI makes a difference between DC11 and DCTERMS (the former is a subset and less precisely specified version of the latter), and the current references mixed the two. The dates were also outdated.

There is a DC-TERMS references in the legacy file, too, I am not sure what to do with that one. It is correct, but the date is old. Is it possible to alias to the new version instead? I am not sure how legacy is used

Cc: @tobie

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -506,9 +506,6 @@
         "date": "14 June 2012",
         "status": "DCMI Recommendation"
     },
-    "dcterms": {
-        "aliasOf": "DCTERMS"
-    },
     "DC11": {
         "authors": [
             "Dublin Core metadata initiative"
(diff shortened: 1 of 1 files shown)


refs/biblio.json
```diff
@@ -497,14 +497,29 @@
...
+    "DC11": {
+        "authors": [
+            "Dublin Core metadata initiative"
+        ],
+        "href": "http://dublincore.org/documents/2012/06/14/dces/",
+        "title": "Dublin Core Metadata Element Set, Version 1.1",
+        "date": "14 June 2012",
+        "status": "DCMI recommendation"
+    },
+    "dc11": {
```
tobie: Not necessary. Case sensitivity is handled automatically.

tobie: Actually, those aliases are what's beaming the tests. You should be good to go once those are fixed.

iherman: Oops, sorry, I did not realize that case sensitivity is handled automatically (which is great, actually:-).

I have made a new commit by removing those.

PR #267, opened 2016-04-13
Added Refs for POE WG

riannella:
Mainly ODRL specs


tobie:
Shipped: http://www.specref.org/?q=odrl21-model

PR #269, opened 2016-04-17
Convert all "C" items of legacy strings to structured data

arronei:
@tobie take a look at CHARSET I left out a bit of text because there isn't a place to put it. Any ideas?
Also COMPUTABLE was hacked together. I think we need something like "section" for that one.


arronei:
Issue #147

refs/biblio.json
```diff
@@ -240,9 +248,27 @@
...
+        "href": "http://www.unicode.org/reports/tr26/",
+        "title": "UTR #26: Compatibility Encoding Scheme For UTF-16: 8-BIT (CESU-8)",
+        "publisher": "Unicode Consortium"
+    },
     "CGM": {
         "aliasOf": "NOTE-cgm"
     },
+    "CHARSETS": {
+        "href": "http://www.iana.org/assignments/character-sets",
+        "title": "Characters sets"
```
tobie: s/Characters sets/Character sets/

tobie:
Sorry, was on my phone when I answered your previous questions. I don't have good answers for your questions for now. I suggest we go ahead and merge those as is.

If people have issues with the changes, they'll reach out.

PR #271, opened 2016-04-18
Add CSS aliases

arronei:
* additional aliases based on the great CSS rename

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -386,8 +386,32 @@
     "CSS-ADAPTATION": {
         "aliasOf": "css-device-adapt"
     },
+    "CSS-ANIMATION-1": {
+        "aliasOf": "css3-animations"
+    },
+    "CSS-BACKGROUND-3": {
+        "aliasOf": "css3-background"
+    },
+    "CSS-COLOR-3": {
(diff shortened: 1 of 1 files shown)


tobie:
Mmm. How come those aren't in there automatically. :-/

tobie:
Mmm, I guess none of those reached WD with their new names yet.

Slightly concerned about what's going to happen once these are merged automatically, though. Will this break stuff? I have no idea.

tabatkins:
Because the webmaster screwed up the publishing requests. :/  Note, for example, that `/TR/css-animations/` works, but not `/TR/css-animations-1/`.

tobie:
Should we attempt an upstream fix, then?

tabatkins:
Yeah, that's the right thing to do.

arronei:
@tobie @tabatkins should we close this PR then?

tobie:
Yeah, if the CSS WG plans to get those fixed on W3C's end, so that <https://www.w3.org/TR/css-color-3/> effectively redirect to <https://www.w3.org/TR/css3-color/>, Specref will get the aliasing for free the next time it parses the RDF files. We also won't incur a maintenance cost that way.

PR #272, opened 2016-04-22
replace http://www.w3.org/ by https://www.w3.org/

AxelNennker:


tobie:
Thanks for looking into this. While I'm all for referencing https refs when these are available, a shotgun approach like that won't work.

For example, some (most?) of your changes to refs/w3c.json will probably get overwritten the next time there's an automatic update. Similarly, data/w3c-specs.txt is external data that was used in an o.d import script. Updating it to https might not necessarily be super useful (maybe deleting it is a better idea). Etc.

Considering what we're trying to achieve here and attempting to fix just that seems like a more solid approach.

AxelNennker:
Feels like turtles all the way down ;-)
I'll undo the w3c-specs.txt changes.
So what needs to be changed in specref so that bikeshed's [biblio](https://github.com/tabatkins/bikeshed/tree/master/bikeshed/spec-data/readonly) stuff uses https?

tobie:
What is it you're trying to achieve, precisely, though?

AxelNennker:
@w3c moved their infrastructure to [https](https://www.w3.org/2001/tag/doc/web-https). So all specs are now available through HTTPS.
While looking at [webauthn](https://github.com/w3c/webauthn) I noticed that the references to the specs are not https. webauthn like many other specs use https://github.com/tabatkins/bikeshed which uses https://github.com/tobie/specref/ 
This chain led me here.

I think that there are some small steps that can be done now like moving data sources like RDF_FILE to https:
https://github.com/tobie/specref/blob/master/scripts/rdf.js#L9
As you wrote above the shotgun method does not work but the small steps should be taken I think.

SpecRef uses https for @IETF refs already. I think SpecRef should take that road for w3c specs too.

tobie:
> I think that there are some small steps that can be done now like moving data sources like RDF_FILE to https:
https://github.com/tobie/specref/blob/master/scripts/rdf.js#L9

So what would that technically change? Make sure that no one tampered with the data source? Also, I'm not sure whether that doesn't require code changes to the script, and/or previous entries.

AxelNennker:
Yes that would **only** make sure that the data source was not tampered with during transit.
No big deal but should be done I think.
The code change is at the bottom of the script. 
This line might need to be changed, I think.
https://github.com/tobie/specref/blob/master/scripts/rdf.js#L311

This discussion should be part of an issue not part of an PR I think...
The PR is too big and because I have put everything into one commit it is now hard to reuse even partially. Should I close it and create an issue?

(How do I test changes locally?)

tobie:
> This discussion should be part of an issue not part of an PR I think...
> The PR is too big and because I have put everything into one commit it is now hard to reuse even partially. Should I close it and create an issue?

Yeah, let's split things up into seperate issues/PRs.

> (How do I test changes locally?)

You can run `npm test`. That said, there are no tests for the scripts, so testing that those behave normally is a slow, manual process. :(

PR #277, opened 2016-04-29
Specref274

AxelNennker:


scripts/get-shortname.js
```diff
@@ -7,6 +7,12 @@ var SPECIAL_CASES = {
     'http://www.w3.org/TR/1998/WD-HTTP-NG-goals': "HTTP-NG-goals",
     'http://www.w3.org/TR/REC-html32': "HTML32",
     'http://www.w3.org/TR/2001/WD-xhtml1-20011004/': 'xhtml1'
```
tobie: Missing a coma here.

scripts/get-shortname.js
```diff
@@ -7,6 +7,12 @@ var SPECIAL_CASES = {
...
+    'https://www.w3.org/Search/9605-Indexing-Workshop/ReportOutcomes/S6Group2': "S6Group2",
+    'https://www.w3.org/TR/1998/NOTE-P3P10-Protocols': "P3P10-Protocols",
+    'https://www.w3.org/1999/05/WCA-terms/': "WCA-terms",
+    'https://www.w3.org/TR/1998/WD-HTTP-NG-goals': "HTTP-NG-goals",
+    'https://www.w3.org/TR/REC-html32': "HTML32",
+    'https://www.w3.org/TR/2001/WD-xhtml1-20011004/': 'xhtml1'
```
tobie: duplicate entry

AxelNennker:
regarding the duplicate entries: the plan was to have get-shortname resolve http://blabla and https://blabla to the same short name. Or am I missing something?

tobie:
No, you're not missing something. I'm just dumb.

tobie:
I would prefer handling this differently, though. Let me look for a solution.

AxelNennker:
other topic: used jslint on the script because that would have caught the missing comma.
jslint is complaining about three other small things
- use strict in function
- braces around script body
I would like to add those. Should I? 
jslint can be annoying but it helps catch me being stupid sometimes.

AxelNennker:
I thought about changing the pattern matching instead of adding to the array but this seemed easier and given the short number of entries in the array is probably not slowing things down too much.

tobie:
I'm sort of hostile to enforcing linting.

AxelNennker:
Don't want to enter hostilities :-)
But enforcing linting is different to accept code style changes that make jslint happy.

While jslinting helper.js jslint complaint that body is not used which made me wonder whether changing the request method from "GET" to "HEAD" might still achieve the goal to check the url in getURL but speed things up because then not the whole TR is fetched...

tobie:
> But enforcing linting is different to accept code style changes that make jslint happy.

Totally fine with that as long as you don't complain if/when it breaks in a while. :)

tobie:
> While jslinting helper.js jslint complaint that body is not used which made me wonder whether changing the request method from "GET" to "HEAD" might still achieve the goal to check the url in getURL but speed things up because then not the whole TR is fetched...

I don't think helper.js is used anywhere, actually.

scripts/helper.js
```diff
@@ -26,7 +27,7 @@ exports.parseURL = parseURL;
 function parseURL(url) {
     var output = null;
     url = url || '';
-    if (url.indexOf(TR_URL) === 0 && url != "http://www.w3.org/TR/1999/WAI-WEBCONTENT-19990505") {
+    if (((url.indexOf(TR_URL) === 0) || (url.indexOf(TR_URL_HTTP) === 0)) && url != "http://www.w3.org/TR/1999/WAI-WEBCONTENT-19990505") {
```
tobie: You probably need to have https://www.w3.org/TR/1999/WAI-WEBCONTENT-19990505 here too. :(

scripts/rdf.js
```diff
@@ -6,7 +6,7 @@ var request = require('request'),
     runner = require('./run'),
     getShortName = require('./get-shortname');
     
-var RDF_FILE = "http://www.w3.org/2002/01/tr-automation/tr.rdf";
+var RDF_FILE = "https://www.w3.org/2002/01/tr-automation/tr.rdf";
```
tobie: This has already been merged.

scripts/rdf.js
```diff
@@ -310,5 +310,5 @@ function _cloneJSON(obj) {
 }
 
 function isGeneratedByThisScript(ref) {
-    return ref.source == "http://www.w3.org/2002/01/tr-automation/tr.rdf" || ref.source == RDF_FILE;
+    return ref.source == "https://www.w3.org/2002/01/tr-automation/tr.rdf" || "http://www.w3.org/2002/01/tr-automation/tr.rdf" || ref.source == RDF_FILE;
```
tobie: So has this.

PR #281, opened 2016-05-11
Add WICG import.

### LINKED ISSUE #266: Can / should WICG specs gets added automatically?
Increasingly new APIs are being specified first in the W3C [WICG](https://github.com/wicg).  I was going to submit a PR to add a couple APIs that are shipping in Chrome (eg. [Intersection Observer](https://github.com/wicg) and [InputDeviceCapabilities](http://wicg.github.io/InputDeviceCapabilities/), but realized it may be better to try to handle this automatically if possible.

reillyeon:
Resolves #266.


tobie:
Thanks a lot for this. There's a lot of duplication between the WHATWG import script and this one. Think you could somehow abstract that, so the script gets called with a couple of params, e.g.:

```terminal
node scripts/import.js [src] [publisher]
```

reillyeon:
That turned out to be a little more complex than anticipated because run.js was doing weird things with command line arguments, but done.

tobie:
Yeah. This software is more duct tape than code. Apologies. :)

scripts/fetch-refs.js
```diff
@@ -1,16 +1,22 @@
...
-var current = runner.readBiblio(FILENAME);
+var SOURCE = process.argv[2];
+var PUBLISHER = process.argv[3];
+var FILENAME = PUBLISHER.toLowerCase() + ".json";
+
+var biblio = helper.readBiblio();
+var current = helper.readBiblio(FILENAME);
 var refs = bibref.createUppercaseRefs(bibref.expandRefs(bibref.raw));
-console.log("Updating WHATWG references...");
+console.log("Updating WICG references...");
```
tobie:
```js
console.log("Updating", PUBLISHER, "references...");
```

tobie:
I'm not sure I understand the rationale for moving stuff from the runner to the helper. Can you explain? (I want to make sure we don;t accidentally break other scripts.)

reillyeon:
run.js is not safe to require from other scripts because it reads its command line arguments. I moved the helper functions other scripts were using into helper.js.

tobie:
OK, that makes sense. Thanks!

tobie:
Thanks!

PR #284, opened 2016-05-28
Update Authoring Practices editors draft URL

MichielBijl:
The APG has been split off into its own repo, so it also has a new ED.

### DIFF
--- a/refs/w3c.json
+++ b/refs/w3c.json
@@ -54698,7 +54698,7 @@
         "rawDate": "2016-03-17",
         "status": "WD",
         "publisher": "W3C",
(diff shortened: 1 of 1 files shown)


tobie:
Unfortunately that would get overwritten pretty much instantly by the update bot.

You have two options here:

1. Update the spec with the new ED URL. publish a new WD. And the update bot will take care of the rest. This is by far the best and most recommended version. 
2. If you can't update the WD for a good reason (silly process issues, etc.), you can go add your new ED URL to the ED dictionary of the [import script](https://github.com/tobie/specref/blob/master/scripts/rdf.js#L42).

MichielBijl:
A new WD will be published in a couple months, so should resolve itself in that case.

PR #297, opened 2016-09-02
Updated odrl-w3c-model. Added odrl-w3c-vocab

riannella:
Updated odrl-w3c-model with new URL and status. 
Added new odrl-w3c-vocab reference

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -2708,11 +2708,26 @@
             "Renato Iannella",
             "Serena Villata"
         ],
-        "href": "http://w3c.github.io/poe/model/WD/",
+        "href": "http://www.w3.org/TR/odrl-model/",
         "publisher": "W3C",
-        "status": "W3C Editors Draft",
-        "rawDate": "2016-06-20",
+        "status": "W3C First Public Working Draft",
+        "rawDate": "2016-07-21",
         "title": "ODRL Information Model"
+    },
+    "odrl-w3c-vocab": {
+        "authors": [
+            "Renato Iannella",
+            "Michael Steidl",
+            "Mo McRoberts",
+            "Stuart Myles",
+            "James Birmingham",
+            "Víctor Rodríguez-Doncel"
+        ],
+        "href": "http://www.w3.org/TR/vocab-odrl/",
+        "publisher": "W3C",
+        "status": "W3C First Public Working Draft",
+        "rawDate": "2016-07-21",
+        "title": "ODRL Vocabulary & Expression"
     },
     "odrl2-req": {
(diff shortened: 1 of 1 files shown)


refs/biblio.json
```diff
@@ -2708,12 +2708,27 @@
             "Renato Iannella",
             "Serena Villata"
         ],
```
tobie: Instead you just want to alias this one to the W3C entry like so:

```json
"odrl-w3c-model": { "aliasOf": "odrl-model" }
```

refs/biblio.json
```diff
@@ -2708,12 +2708,27 @@
...
-        "href": "http://w3c.github.io/poe/model/WD/",
+        "href": "http://www.w3.org/TR/odrl-model/",
         "publisher": "W3C",
-        "status": "W3C Editors Draft",
-        "rawDate": "2016-06-20",
+        "status": "W3C First Public Working Draft",
+        "rawDate": "2016-07-21",
         "title": "ODRL Information Model"
     },
+    "odrl-w3c-vocab": {
```
tobie: This one already exists in the repository as [[vocab-odrl](http://www.specref.org/?q=vocab-odrl)].

riannella: I could not see the "vocab-odrl" entry in biblio.json?

refs/biblio.json
riannella: No alias is needed. odrl-w3c-model is what we use

tobie: You still have to turn it into an alias.

refs/biblio.json
tobie: It's in the W3C file.

riannella: Sorry, I wasn't aware that someone had added these two to the w3c file. Should all W3C specs be in there instead of biblio?

tobie: All W3C specs published on /TR are automatically added to this file which is regenerated every hour. It can't be manually edited.

tobie: Legacy W3C entries which are in biblio should just be aliased to their automatically generated entry in the W3C file.

riannella: Ok, I see now ;-) 
So we should use the reference in the w3c file in our TR specs?
Then disregard the addition of odrl-w3c-vocab to the bibilo and add an alias from odrl-w3c-model to the w3c version

tobie: Precisely.

tobie: Will you be submitting a new PR or editing this one?

riannella: Just edited it

Issue #305, opened 2016-10-27
add 'FIPS-186-4'


equalsJeffH:
bikeshed just told me:
```
FATAL ERROR:\033[0m Couldn't find 'FIPS-186-4' in bibliography data. Did you mean:
  fips-186-3
  fips-180-4
  fips-180-3
  cwg1864
  p0186r0
Did not generate, due to fatal errors
```
FIPS-186-4 is here: http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf

tobie:
References are pretty easy to add: https://github.com/tobie/specref#manual-changes

You could just add it below the [FIPS-180-4](https://github.com/tobie/specref/blob/ee2359a/refs/biblio.json#L993-L997) reference.

equalsJeffH:
was unable to submit patch right away, so sub'd issue so I/we wouldn't fergetabboudit.

PR #306, opened 2016-10-28
update [DSS] to point at FIPS 186-4

equalsJeffH:
[DSS] entry already refs FIPS 186-3, so just updated that to -4.


equalsJeffH:
fixes #305

tobie:
And deployed: http://www.specref.org/?q=FIPS%20PUB%20186-4

Thanks for the support.

PR #314, opened 2016-12-08
Adding second author to image-capture spec

miguelao:


tobie:
That should be automatic the next time the spec gets published on /TR. 

Whatever you fix here would get overwritten nearly immediately by the sync bot.

miguelao:
That'd work for me, but I have been co-editor since ~June and the authors list hasn't updated...

tobie:
Have you submitted a new WD or are you planning to soon?

miguelao:
IIUC we're working now against `gh-pages` branch and do no timed releases, @dontcallmedom keep me true please.

tobie:
Sounds like an upstream issue. Cc'ed you in the email I just sent.

PR #319, opened 2017-01-03
feat: add timestampinfo to requests

marcoscaceres:


marcoscaceres:
rebasing...

index.js
```diff
@@ -18,6 +18,8 @@ app.use(require("errorhandler")(errorhandlerOptions));
 // bibrefs
 app.get('/bibrefs', function (req, res, next) {
     var refs = req.query["refs"];
+    res.setHeader("Expires", new Date(Date.now() + 86400000).toUTCString());
```
tobie: Why 24h? Fwiw, the update script runs every hour.

index.js
```diff
@@ -9,7 +9,7 @@ if (process.env.NODE_ENV == "dev" || process.env.NODE_ENV == "development") {
     errorhandlerOptions.dumpExceptions = true;
     errorhandlerOptions.showStack = true;
 }
-
+app.enable("etag");
```
tobie: Given the app is restarted every time there is a change in the DB, we're not going to get anything out of this unless the etags are stored somewhere. If so, where?

index.js
marcoscaceres: 24 hours seems like a fairly sane default per spec, no?

index.js
tobie: Oh, wait. They're content hashes, no? So that should work.

marcoscaceres: Yeah, I think they are content hashes: 
https://github.com/expressjs/express/blob/3ffceff3ed356a711fce5fee6ed00a03f43affd8/lib/utils.js#L21

index.js
tobie: Yeah, It's rare that we get manual update requests nowadays, so I guess the use-case where someone just fixed the db and wants to see the update right away in their spec is rare enough.

tobie:
Alright, this LGTM. Let's give it a spin. Mind messaging spec-prod to let them know we've shipped this?

marcoscaceres:
> Mind messaging spec-prod to let them know we've shipped this?

Will do! Thanks! 

Will this be deployed straight away? I'm cooking up some new goodness based on this :)

marcoscaceres:
YAY! I can see the 304 in https://specref.herokuapp.com/bibrefs?refs=dom 💃

tobie:
Yup, and the proper cache headers are there too. Awesome!

PR #324, opened 2017-01-16
feat(biblio): add WebIDL-LS

marcoscaceres:


marcoscaceres:
Because we can't have enough links to WebIDL 👍

marcoscaceres:
Went with "WebIDL-LS" instead of L2. Also set status to "ED".

tobie:
Frankly, I'm torn, here.

On one hand, I think WebIDL href should just point to the ED.
On the other I see @marcoscaceres's point, in which case I think WebIDL-2 should be an alias of WebIDL-LS.

marcoscaceres:
Problem is that WebIDL-2 is already pointing to the wrong place (aliasOf: webIDL). Maybe no one is using that though and we can just take it over. I'm ok with correcting WebIDL-2 to point to the ED. Let's just do it.

tobie:
OK, fixed.

Waiting for Travis, then deploying.

marcoscaceres:
Awesome, thanks for fixing.

PR #328, opened 2017-01-20
Alias URL to WHATWG-URL.

tobie:


darobin:
OMG, and the world didn't even end!

tobie:
@darobin it's actually been non-trivial to get this not to be overwritten while preserving automatic updates.

PR #330, opened 2017-02-16
refactor: structure some ISO entries out of legacy

marcoscaceres:
* includes iso8601 and iso3166


marcoscaceres:
@tobie, sorry, what's the fix here?

marcoscaceres:
ah, we just the author :)

PR #345, opened 2017-04-05
Update some UAX entries

frivoal:
Points UAX* bibliographic entries to their latest incarnation,
updating relevant metadata (authors, date, title) accordingly.

Not included in this commit are those that have been superseded (UAX21
& UAX27). See https://github.com/w3c/csswg-drafts/issues/1169#issuecomment-291779023

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -186,9 +186,9 @@
             "Aharon Lanin",
             "Andrew Glass"
         ],
-        "href": "http://www.unicode.org/reports/tr9/",
+        "href": "http://www.unicode.org/reports/tr9/tr9-35.html",
         "publisher": "Unicode Consortium",
-        "rawDate": "2014-06-05",
+        "rawDate": "18 May 2016",
         "status": "Unicode Standard Annex #9",
         "title": "Unicode Bidirectional Algorithm"
     },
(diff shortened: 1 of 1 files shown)


littledan:
Can we just point to the latest version, at index.html? It looks like we already do that for UTR 26, UAX 9, and the main Unicode standard.

frivoal:
I don't know. Should we? If we do that, the date and list of authors gets out of sync with the document being pointed to. It seems to me that this should be solved by automating the update, rather than by pointing do a moving target. But I am not sure.

littledan:
Even though CSS has been pointing to these old versions, Chromium has been updating to the newest version on a regular basis. I think a better way to encourage implementers to upgrade is, when something useful comes in a new Unicode version, to add a web platform test that depends on this behavior.

frivoal:
+1 to adding a test that depends on the newest spec every time there's one.
I am also supportive of trying to keep bibliographic data up to date and pointing to the latest version of things, but I just wonder what the best way to do it is.

I can of course update this PR to point the the ever-green link, but was under the impression that this was not the way things should be done. @tobie, any feedback?

tobie:
Well, there are two thing you want to avoid:

1. pointing to the wrong document,
2. providing metadata that's out of sync with the document you point to, as it risks confusing implementors on what your intentions are.

So the ideal solution is of course to automate this. If that's not possible, then you want to keep manually updating it, and if you can't trust yourself with that, maybe don't include author or publications dates.

frivoal:
I cannot promise to be always on the top of things, but updating it more often than the 10+years that this was stalled for sound doable. I suppose once I do it once or twice, I'll be bite the bullet and automate it. But for now I just want a quick fix, as I am neither the editor of the document being pointed to nor the one doing the pointing, just a consumer of the output and have many more things on my plate.

tobie:
Sounds good to me.

refs/biblio.json
```diff
@@ -186,9 +186,9 @@
             "Aharon Lanin",
             "Andrew Glass"
         ],
-        "href": "http://www.unicode.org/reports/tr9/",
+        "href": "http://www.unicode.org/reports/tr9/tr9-35.html",
         "publisher": "Unicode Consortium",
-        "rawDate": "2014-06-05",
+        "date": "18 May 2016",
```
tobie: fyi, `rawDate` gets converted to `date` by the app. It's just more machine readable.

tobie:
Thanks for your contribution!

Thanks again!

frivoal:
@tobie Thanks for granting access. I'll use it responsibly.

As a follow up to this PR, what do you think of the question I raised in https://github.com/w3c/csswg-drafts/issues/1169#issuecomment-291779023 ?

tobie:
@frivoal: Does https://github.com/w3c/csswg-drafts/issues/1169#issuecomment-291801382 answer your questions?

littledan:
I'm wondering why it makes sense to have some references in this document versioned and some unversioned.

tobie:
In which document?

littledan:
In this biblio file, the latest version of Unicode is referenced, and that is used together with these references in CSS Text Level 3.

tobie:
I'm not sure what you mean. :-/

PR #351, opened 2017-04-14
Mark UAX21 and UAX27 as obsolete

### LINKED ISSUE #347: UAX21 and UAX27 need updating — but how?
The UAX21 and UAX27 entries are out of date. Not only are there later incarnations of the same documents, but also these documents have eventually been discontinued and their content merged to an other spec (Core Specification of the Unicode Standard). What should we do?

It makes sense to me to:
1. Update them to the latest standalone version
2. add an `obsoletedBy` key to point to the document that replaces them.

However, client software that read from specref do not currently support `obsoletedBy`. Should we just fix them? Should we mark it some other way?

--

issue raised from https://github.com/w3c/csswg-drafts/issues/1169#issuecomment-291779023
bikeshed issue about supporting `obsoletedBy`: https://github.com/tabatkins/bikeshed/issues/427

frivoal:
Closes #347

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -2178,7 +2178,8 @@
         "title": "Case Mappings",
         "date": "26 March 2001",
         "status": "Unicode Standard Annex #21",
-        "publisher": "Unicode Consortium"
+        "publisher": "Unicode Consortium",
+        "obsoletedBy": [ "UNICODE" ]
     },
     "UAX24": {
         "authors": [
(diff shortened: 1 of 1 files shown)


tobie:
LGTM. Please don't close #347, though, until the list of problematic ids that Tab mentioned is fixed or gets its own issue.

frivoal:
This should be a mundane change, but it's the first one of its kind (using "obsoletedBy"), so asking for review anyway.

frivoal:
I've opened a separate issue for the problem reported by Tab: https://github.com/tobie/specref/issues/352

Merging.

PR #353, opened 2017-04-14
Convert the legacy MDN SETCAPTURE ref

siusin:
Mouse Capture is unspecified and some W3C specs (f.ex. [Pointer Lock](https://w3c.github.io/pointerlock/#bib-MSDN-SETCAPTURE)) are using it as an informative ref.


refs/biblio.json
```diff
@@ -3010,5 +3010,14 @@
...
         "publisher": "W3C Web Bluetooth Community Group"
+    },
+    "mdn-setcapture": {
+        "authors": [
+            "Shepherd, Eric; et. al"
+        ],
+        "href": "https://developer.mozilla.org/en-US/docs/Web/API/Element/setCapture",
+        "title": "Element.setCapture()",
+        "publisher": "Mozilla Developer Network",
+        "rawDate": "2017-04-05"
```
tobie: I think you might as well drop the date here altogether.

siusin: Make sense. How about the authors?

tobie:

PR #357, opened 2017-05-08
add HTML extensions draft

chaals:
This is a Work in Progress, and hopefully will move to a W3C publication that gets picked up automatically at some point...


tobie:
Thanks, @chaals!

PR #364, opened 2017-06-22
Remove legacy reference to MIDI

svgeesus:


svgeesus:
@tobie Hi Tobie. Travis won't let me remove the legacy MIDI ref as it says it is "maintained" (but the link is broken in the existing ref) and it won't let me add the new ref (with the right link, publisher, etc) because it conflicts with the legacy ref.
WebMIDI won't build with Respec because "legacy specrefs are no longer supported" so it has no refs section at all.
Advice?

tobie:
Yeah, you need to combine both PRs in a single one and all should be good (though the test suite might balk at your date format (try using the rawDate prop instead) and the "edition" prop (you might need to remove that).

tobie:
FYI, the test suite verifies that no reference is removed in a PR, which is why you need to combine both PR to pass the test suite. This is what "Previous references are maintained" and "Reference [[MIDI]] is maintained" refer to (they're the name of the test suite and the specific test respectively). This has nothing to do with whether the reference itself is maintained or not (which would be kind of hard for a test suite to figure out, tbh).

Agree the wording here could be improved. Suggestions welcomed.

PR #365, opened 2017-06-29
Convert legacy INBANDTRACKS ref

bbcrddave:
https://github.com/w3c/webmediaapi is informatively referencing the INBANDTRACKS specification, and we'd like to do that ReSpec-error-free :smile:


tobie:

PR #367, opened 2017-07-12
Add aliases for all the 'correct' CSS spec names known by Shepherd, b…

tabatkins:
…ut not yet known in the W3C systems.


tobie:
If these aliases are going to end up in W3C data, it might be more future-proof to put them all in w3c.json instead of biblio.json.

tabatkins:
Isn't that auto-generated?

tabatkins:
(There's already several of these CSS aliases in biblio.json anyway.)

tobie:
It _is_, but it keeps existing data unless it gets overwritten.

tobie:
Yeah, I know we're not super consistent here. My concern is essentially that there might be conflicts once those get added to W3C data. This won't hapen if the data's in w3c.json. It _might_ if it's in biblio.json.

tobie:
Not a deal blocker anyway. Feel free to merge as is.

tabatkins:
Ah, kk. Will those conflicts end up showing up in your analytics or something, so we can remove them?

tabatkins:
(I can move all the CSS-related ones over to w3c.json in another PR if you think that's better.)

tobie:
If it's _really_ an issue (Specref is 90% duck tape, so I'm never _sure_ whether something will break and thus err on the conservative side), we'll find out once W3C starts adding those in its tr.rdf file; the updating process will fatal and we'll just fix it then.

tabatkins:
Works for me. ^_^

PR #369, opened 2017-08-03
Document where W3C/IETF import scripts are

strugee:
I couldn't find this myself, so I figured I would document it for others. `rdf.js` in particular isn't very well-named.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,6 @@
+# Auto-update scripts
+
+Here are the important files in this directory:
+
+* `rfc.js` merges IETF data
+* `rdf.js` merges W3C data


tobie:
Yeah. There's a lot of legacy stuff lying around.

We could also just rename the files, no?

strugee:
@tobie I'd be happy to do that, just wasn't sure what other stuff would break because it expected those filenames

tobie:
@strugee awesome.

I think if you just rename `rdf.js` -> `w3c.js` and `rfc.js` -> to `ietf.js` and fix both those references in `udate-all`, you should be good. :crossed_fingers:

tobie:
There's a bunch more scripts in `scripts/fetch-helpers`, btw.

Also curious as to what brought you to check out Specref's innards, if you're able to share.

strugee:
> Also curious as to what brought you to check out Specref's innards, if you're able to share.

I wanted to write a script to automatically pull in specs from http://microformats.org, but it turned out to be too complicated to parse automatically and they don't change that often anyway. So #370 happened instead :)

strugee:
Closing and will submit a new PR with the rename.

PR #370, opened 2017-08-03
Add microformats2 specifications

strugee:
/cc @tantek to make sure this is all right.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -1205,6 +1205,116 @@
         "rawDate": "2015-07",
         "status": "Working Draft"
     },
+    "h-adr": {
+        "authors": [
+            "Tantek Çelik"
+        ],
+        "href": "http://microformats.org/wiki/h-adr",
+        "title": "h-adr",
+        "publisher": "microformats.org",
+        "status": "Draft Specification"
+    },
+    "h-card": {
+        "authors": [
+            "Tantek Çelik"
+        ],
+        "href": "http://microformats.org/wiki/h-card",
+        "title": "h-card",
+        "publisher": "microformats.org",
+        "status": "Draft Specification"
+    },
+    "h-entry": {
+        "authors": [
+            "Tantek Çelik"
+        ],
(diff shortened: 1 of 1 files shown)


tantek:
Better to use "status": "Living Specification" for all of them from an external perspective. Internally see http://microformats.org/wiki/process#Moving_from_Stage_to_Stage for draft vs spec vs standard if you're curious about the different levels according to microformats process. All of those are "living specs" as it were.

tobie:
This LGTM.

Feel free to merge once you've addressed the changes requested by @tantek.

strugee:
Pushed c14dfab05a2722fe4a17688a401c207c59861906 to address the status issue. Thanks for the write access :thumbsup: :tada:

tobie:
@tantek, fwiw, the WHATWG folks tend to use the term "Living standard." I don't care either way.

strugee:
@tobie based on http://microformats.org/wiki/process#Standard I guess the microformats folks differentiate between "specification" and "standard"

tobie:
@strugee please don't _ever_ merge with checks pending.

Please also squash merges unless there's a god reason not to (e.g. preserve history).

strugee:
@tobie ah sorry, I read that guideline in the README but figured it was so trivial there was no need to wait for Travis. I'll be more careful in the future. Should I force-push a squash for this merge?

tobie:
> I read that guideline in the README but figured it was so trivial there was no need to wait for Travis.

A typo in the JSON and the deploy script just stalls until I go manually fix the problem.

>  Should I force-push a squash for this merge?

No, please don't.

tobie:
Don't worry about it, though. :)

tantek:
@tobie had not noticed that difference before, and the mf2 specs are consistently using "living specification" currently so let's stick with that for now. I'm open to switching to "living standard", no strong feelings there. Happy to make it an open question in the microformats community and have folks discuss it to see if there is a consensus either way.

PR #374, opened 2017-08-15
Update reference to [COLORIMETRY]

frivoal:
@svgeesus, your review is appreciated.


svgeesus:
Sorry to have missed this while on vacation but yes, correct.
Also btw http://www.cdvplus.cz/file/3-publikace-cie15-2004/

PR #375, opened 2017-09-08
Add keys and patterns for ISO and OGC standard numbers 

dr-shorthair:
(similar to IETF RFC numbers)
to support upcoming submission of a set of ISO and OGC specs to the /refs/

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -1331,35 +1331,40 @@
         "title": "Information technology — Coding of audio-visual objects — Part 30: Timed text and other visual overlays in ISO base media file format",
         "href": "http://www.iso.org/iso/home/store/catalogue_tc/catalogue_detail.htm?csnumber=63107",
         "date": "11 March 2014",
-        "publisher": "ISO/IEC"
+        "publisher": "ISO/IEC",
+        "isoNumber": "ISO 14496-30:2014"
     },
     "ISO3166": {
         "date": "2013",
         "href": "http://www.iso.org/iso/catalogue_detail?csnumber=63545",
         "publisher": "International Organization for Standardization (ISO)",
         "status": "ISO 3166-1:2013",
-        "title": "ISO 3166: Codes for the representation of names of countries and their subdivisions."
+        "title": "ISO 3166: Codes for the representation of names of countries and their subdivisions.",
+        "isoNumber": "ISO 3166-1:2013"
     },
     "ISO4217": {
         "date": "2015",
         "href": "http://www.iso.org/iso/home/standards/currency_codes.htm",
         "publisher": "ISO",
         "status": "International Standard",
-        "title": "Currency codes - ISO 4217"
+        "title": "Currency codes - ISO 4217",
+        "isoNumber": "ISO 4217:2015"
     },
     "ISO8601": {
         "href": "http://www.iso.org/iso/catalogue_detail?csnumber=40874",
         "publisher": "International Organization for Standardization (ISO)",
         "date": "2004",
         "status": "ISO 8601:2004",
-        "title": "Representation of dates and times. ISO 8601:2004."
+        "title": "Representation of dates and times. ISO 8601:2004.",
+        "isoNumber": "ISO 8601:2004"
     },
     "ISOBMFF": {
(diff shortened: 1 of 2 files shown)


dr-shorthair:
Will also need to design scrpts iso.js and ogc.js to use the new fields in formatted output.

tobie:
I was wondering if there was a reason to have separate identifiers fields rather than a unique one. I guess it depends on what the use cases for them is. Any thoughts on that front?

dr-shorthair:
I was trying to move forward non-disruptively, and when I spotted the rfcNumber pattern it looked like a clonable capability. There is also the validation issue - each of these numbering systems has its own, relatively predictable pattern which can be validated with a regex which is probably a useful check. 

FWIW - I have a good relationship with OGC and can almost certainly get them to host an ogc.json file for you to harvest regularly. 

I'm not as close to ISO/TC 211 - who are responsible for the ISO 191NN series (around 50 standards) - but a colleague at the EC has already prepped the json for a key subset of these. They are important inputs to INSPIRE documentation. 

And standards from both of these series have been referred to in recent standards-track documents from W3C - see  https://www.w3.org/TR/owl-time/#references and https://www.w3.org/TR/vocab-ssn/#references 

So I think there is probably merit in them being added to Specref for ongoing use.

tobie:
> I was trying to move forward non-disruptively, and when I spotted the rfcNumber pattern it looked like a clonable capability. There is also the validation issue - each of these numbering systems has its own, relatively predictable pattern which can be validated with a regex which is probably a useful check.

That sounds good!

> FWIW - I have a good relationship with OGC and can almost certainly get them to host an ogc.json file for you to harvest regularly.

That would be ideal.

> I'm not as close to ISO/TC 211 - who are responsible for the ISO 191NN series (around 50 standards) - but a colleague at the EC has already prepped the json for a key subset of these. They are important inputs to INSPIRE documentation.

Happy to help modify the polling scripts if you need help with this.

> And standards from both of these series have been referred to in recent standards-track documents from W3C - see https://www.w3.org/TR/owl-time/#references and https://www.w3.org/TR/vocab-ssn/#references
>
> So I think there is probably merit in them being added to Specref for ongoing use.

Agreed.

tobie:
If you want the data exposed (and not just stored in the JSON files), you'll need to:

1. update the [processed reference schema](https://github.com/tobie/specref/blob/master/schemas/reference.json), and
2. add the property names you want exposed to the [PROPS array](https://github.com/tobie/specref/blob/master/lib/bibref.js#L6-L24).

You'll notice that we haven't exposed the rfcNumber so far. Unless you have a use case for this data being exposed, I'd rather it wasn't. As this gives us the ability to eventually expose all of this info as "publisherIdentifier" (or equivalent) if it turns out having that data grouped as such is better.

dr-shorthair:
> Unless you have a use case for this data being exposed, I'd rather it wasn't. 

For both ISO and OGC the code is not part of the URL nor the human-friendly title, but is considered an important part of the document identifier. For example, ISO requests that its documents are cited something like 

ISO 19156:2011 Geographic Information - Observations and Measurements. International Standard, Geneva. 2011 

etc etc

Now we can't pander to every publishers requirements, but the publisher identifier is clearly important in some cases - many ISO standards are known colloquially by their number (e.g. 8601) though it doesn't appear in the title. 

> As this gives us the ability to eventually expose all of this info as "publisherIdentifier" (or equivalent) if it turns out having that data grouped as such is better. 

Yeah. There clearly is a grouping, though pattern validation is easier with distinct tags. I'll see if I can find a JS person to help. (I'm not a programmer.)

tobie:
> Yeah. There clearly is a grouping, though pattern validation is easier with distinct tags. I'll see if I can find a JS person to help. (I'm not a programmer.)

JSON schema let's you do this with the [oneOf construct](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html#oneof).

tobie:
I decided I'd just merge this in now and we can see how we best want to expose those down the road.

tobie:
Thanks for your contribution!

Thanks again!

dr-shorthair:
> You've thus been granted commit access to the repo. 

Thanks Tobie. Is the expectation that I would now stop working in my fork, and branch here instead?

PR #377, opened 2017-09-12
Add ISO 19100 refs

dr-shorthair:
Initial selection from ISO 19100 series. 
Thanks to @andrea-perego for suggesting


dr-shorthair:
Does this JSON file needed to be added to an inventory so it gets included in specref?

dr-shorthair:
Oh dear - test failing, but does not appear to be in my new content :-(

refs/iso19100.json
```diff
@@ -0,0 +1,293 @@
+{
+    "iso-19101-1": {"aliasOf": "iso-19101-1-2014"},
+    "iso-19101-1-2014": {
+        "authors": ["ISO/TC 211"],
+        "href": "https://www.iso.org/standard/59164.html",
+        "title": "Geographic information -- Reference model -- Part 1: Fundamentals",
+        "publisher": "ISO",
+        "deliveredBy": [{
```
tobie: I'd forgo deliveredBy altogether unless you have a use case for it.

dr-shorthair: The "deliveredBy" model is different in reference.json (list of url, shortname pairs) compared with raw-reference.json (list of URLs)  - is this the problem?

Our intention was to provide the link to the particular ISO Technical Committee which is responsible for all the ISO 19100 series standards. I think that is what deliveredBy is for?

refs/iso19100.json
```diff
@@ -0,0 +1,249 @@
+{
```
tobie: You want to put all manual entries in biblio.jon for now.

dr-shorthair:
Re-based. 
Copied iso-191** refs into biblio.json (for now)

PR #386, opened 2017-09-18
Link to actual webrtc ED

tobie:
Bandaid for w3c/webrtc-pc#942.

### DIFF
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -7,7 +7,10 @@
     },
     "WebCryptoAPI": {
         "edDraft": { "replaceWith": "https://w3c.github.io/webcrypto/Overview.html" }
-    },    
+    },
+    "webrtc": {
+        "edDraft": { "replaceWith": "https://rawgit.com/w3c/webrtc-pc/master/webrtc.html" }
+    },
     "service-workers-1": {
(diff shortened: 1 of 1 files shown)


foolip:
I would rather leave this broken and keep accumulating examples of people being confused until it's fixed properly.

tobie:
> I would rather leave this broken and keep accumulating examples of people being confused until it's fixed properly.

Well it seems the WG doesn't want to fix it for bogus reasons. I agree this fix isn't optimum, but it demonstrates how bogus the arguments are while also improving the situation.

foolip:
The github.io copy is linked from all over the place, at least I don't much feel like updating links that don't come from specref, like our own IDL files.

This will take a long time to propagate to specs that don't change often, and then when the real problem is fixes it'll take time to revert back everywhere.

This is technically correct so I can't object, but I think the problem is just that no one in particular is responsible for sorting out the issue, so more poking will be needed.

tobie:
> This is technically correct so I can't object, but I think the problem is just that no one in particular is responsible for sorting out the issue, so more poking will be needed.

The purpose of Specref isn't to be technically correct, but to support the work around Web standards. If this hurts more than it helps, then let's not do it.

tobie:
Closing as there doesn't appear to be much consensus, here.

PR #387, opened 2017-09-20
Add ISO7812-1

marcoscaceres:
Used by payments wg

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -1367,7 +1367,7 @@
         "date": "1 January 2017",
         "href": "https://www.iso.org/standard/70484.html",
         "publisher": "ISO/IEC 7812-1: Identification cards - Identification of issuers - Part 1",
-        "status": "ISO 7812-1:2017",
+        "isoNumber": "ISO 7812-1:2017",
         "title": "Identification cards - Identification of issuers - Part 1: Numbering system"
     },
     "ISO8601": {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
we should eventually make a "iso.json" file.... biblio.json is becoming quite the chunky monkey.

marcoscaceres:
Didn't know that was a thing! Thanks.

dr-shorthair:
Yes - that is the plan. I added "isoNumber" last week to the schema in anticipation. And then we would need scripts/iso.js to handle it  - but that is beyond my skillset I'm afraid.

marcoscaceres:
> scripts/iso.js to handle it - but that is beyond my skillset I'm afraid.

What would it do? Maybe I can help?

tobie:
> Didn't know that was a thing! Thanks.

It's not reflected in the API for now, FYI. Would this be something you'd want to expose in ReSpec?

marcoscaceres:
Depends... I don't yet understand the value of it over "status". What would be the benefit to readers over what they get today? This is really a question for @dr-shorthair

marcoscaceres:
Btw, can we merge this and then do all the ISO specs at once? I'm keen to get the spec that needs this out the door. Then I can also go back to focusing on improving ReSpec.

tobie:
Sure, just please fix the status and we're good.

marcoscaceres:
So, as ReSpec doesn't support `isoNumber` yet, I added both status and isoNumber... otherwise, there would be a gap when displaying the reference (because it would lack a status).

tobie:
Feel free to merge once green

PR #394, opened 2017-09-29
Fix titles of BackgroundSync and WebUSB

foolip:
The documents themselves don't have "specification draft" it in the titles.


foolip:
Bah. I just went to a place that was in my history without checking what it was. I'll do this PR in the right place.

PR #396, opened 2017-10-01
Connect the EDs for a bunch of css3-* specs

foolip:
Versioned links are used to be pedantically correct.

### DIFF
--- a/refs/w3c.json
+++ b/refs/w3c.json
@@ -21442,7 +21442,8 @@
             "https://www.w3.org/Style/CSS/members"
         ],
         "rawDate": "2013-02-19",
-        "source": "https://www.w3.org/2002/01/tr-automation/tr.rdf"
+        "source": "https://www.w3.org/2002/01/tr-automation/tr.rdf",
+        "edDraft": "https://drafts.csswg.org/css-animations-1/"
     },
     "css3-background": {
         "authors": [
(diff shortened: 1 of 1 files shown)


foolip:
This is to help me connect some more specs to tests, see https://github.com/foolip/day-to-day/commit/2cbf8e2e090207879db98dfdf46c70304a43b1fd#diff-a6a871ecfe441a676b24e4e8d7a231c0R60

TimothyGu:
Also see https://heycam.github.io/webidl/#biblio-css3-conditional.

foolip:
@TimothyGu, are you keeping track of all /TR/ links, or why did you know about that one?

TimothyGu:
@foolip No, I added that ref to Web IDL. I wondered (for a long time) why it was linking to the CR despite the "current" biblio setting in Bikeshed. Thanks for doing this.

foolip:
Ah, I see. I thought maybe you were on a secret mission against /TR/, which I would support :)

tobie:
Are those just missing from tr.rdf?

If so, you're better of adding them manually to w3c.json (I know none of this makes any sense, but…)

tobie:
Basically, overwrites, is for "willful violations", while missing stuff can just go directly into the corresponding ref files.

foolip:
Yes, they're missing from tr.rdf, and after one successful and one still pending attempt to get it updated, I'm not very keen on doing more of that. Will add to w3c.json, which is indeed very confusing :)

foolip:
@tobie, OK, tried again, commit message still works. How do I know that this'll actually work and not get overwritten?

tobie:
> How do I know that this'll actually work and not get overwritten?

Iirc it should only get overwritten once actual ED are added in tr.rdf.

PR #398, opened 2017-10-02
Use HTTPS for www.unicode.org

annevk:


annevk:
Not really sure this is the right place. There's also some references in legacy that could be updated I think.

tobie:
This is good.

tobie:
You can also fix the legacy stuff if you want.

PR #401, opened 2017-10-02
Align whatwg.json with upstream

annevk:


tobie:
So one of the core tenets of Specref is that entries are kept around, you could add an `obsoleteBy` prop instead.

refs/whatwg.json
```diff
@@ -257,7 +257,7 @@
         "aliasOf": "LOADER"
     },
     "WHATWG-MEDIASESSION": {
-        "aliasOf": "MEDIASESSION"
+        "obsoletedBy": "MEDIASESSION"
```
tobie: `obsoletedBy` expects an array, see [the JSON schema](https://github.com/tobie/specref/blob/master/schemas/raw-reference.json#L202-L204). You'll also need to [provide at least a title](https://github.com/tobie/specref/blob/master/schemas/raw-reference.json#L261).

tobie:
I know figuring out how Specref works isn't straightforward. But there's little I can do given the time I have to work on it, and the large overhaul it would need to be more functional.

annevk:
No worries at all.

@foolip what do you think? Leave WHATWG-MEDIASESSION as-is? I'm starting to think you should simply not have removed the original entry in https://github.com/tobie/specref/commit/ef1e98640522cccafbf2786d01ab1fa046d707da and we should have marked that one as `obsoletedBy`.

tobie:
>  I'm starting to think you should simply not have removed the original entry in ef1e986 and we should have marked that one as obsoletedBy.

Yeah, that's generally your best bet.

foolip:
Going with obsoletedBy seems fine. But making the longer an uglier name the primary one seems bad, so I'm not sure I regret moving MEDIASESSION.

tobie:
> But making the longer an uglier name the primary one seems bad, so I'm not sure I regret moving MEDIASESSION.

That's irrelevant, really. All API consumers need to resolve aliases anyway, so it should be completely transparent to the end user.

The reason for prefixing the terms are to avoid conflicts and allow some fluidity for specs mcong from one org to another.

foolip:
So, do you mean that shortest and prettiest name should always been an alias of an org-prefixed name, or that we should make it so once a spec is moved from one org to the other?

I doubt I'll need to fix an issue like this again, so as long as this ends up in a shape that you're all happy with, I'm also happy.

tobie:
> So, do you mean that shortest and prettiest name should always been an alias of an org-prefixed name, or that we should make it so once a spec is moved from one org to the other?

So this would have been ideal but a whole bunch of people complained about it, so we have this kind of hybrid model that works reasonably well as long as you're ready to accept its oddities. 

> I doubt I'll need to fix an issue like this again, so as long as this ends up in a shape that you're all happy with, I'm also happy.

I don't really care personally. I just don't want to spend more time on Specref than necessary unless someone actually wants to pay for my time doing so.

PR #403, opened 2017-10-03
Fix the ED links for Encrypted Media and Touch Events

foolip:
These are stale in https://www.w3.org/2002/01/tr-automation/tr.rdf

### DIFF
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -15,7 +15,7 @@
         "edDraft": { "replaceWith": "https://w3c.github.io/ServiceWorker/" }
     },
     "touch-events": {
-        "edDraft": { "replaceWith": "https://w3c.github.io/uievents/" }
+        "edDraft": { "replaceWith": "https://w3c.github.io/touch-events/" }
     },
     "uievents": {
         "edDraft": { "replaceWith": "https://w3c.github.io/uievents/" }
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -8,8 +8,14 @@
     "WebCryptoAPI": {
         "edDraft": { "replaceWith": "https://w3c.github.io/webcrypto/Overview.html" }
     },
+    "encrypted-media": {
+        "edDraft": { "replaceWith": "https://w3c.github.io/encrypted-media/" }
+    },
     "service-workers-1": {
         "edDraft": { "replaceWith": "https://w3c.github.io/ServiceWorker/" }
+    },
+    "touch-events": {
+        "edDraft": { "replaceWith": "https://w3c.github.io/uievents/" }
     },
     "uievents": {
         "edDraft": { "replaceWith": "https://w3c.github.io/uievents/" }


overwrites/w3c.json
```diff
@@ -8,9 +8,15 @@
...
         "edDraft": { "replaceWith": "https://w3c.github.io/webcrypto/Overview.html" }
     },
+    "encrypted-media": {
+        "edDraft": { "replaceWith": "https://w3c.github.io/encrypted-media/" }
+    },
     "service-workers-1": {
         "edDraft": { "replaceWith": "https://w3c.github.io/ServiceWorker/" }
     },
+    "touch-events": {
+        "edDraft": { "replaceWith": "https://w3c.github.io/uievents/" }
```
tobie: You mean: https://w3c.github.io/touch-events/, right?

tobie:
So just FYI, the downside of using overwrites is that you need to manually fix those if the data changes.

overwrites/w3c.json
foolip: Oops, of course!

foolip:
> So just FYI, the downside of using overwrites is that you need to manually fix those if the data changes.

Do you mean if the edDraft value changes in tr.rdf, or if the name of the entry as a whole changes? I could survive either. The difficulty of patching tr.rdf makes me do it this way.

@sideshowbarker, is this a problem that could be resolved W3C-side by putting tr.rdf on GitHub? I've sent patches by email, but it's just too much overhead to keep track of.

tobie:
> Do you mean if the edDraft value changes in tr.rdf, or if the name of the entry as a whole changes? I could survive either.

Both.

> The difficulty of patching tr.rdf makes me do it this way.

Yeah, that's basically just shifting the maintenance burden to Specref (i.e. to me), so I tend to favor upstreaming stuff).

> @sideshowbarker, is this a problem that could be resolved W3C-side by putting tr.rdf on GitHub? I've sent patches by email, but it's just too much overhead to keep track of.

iirc @deniak is staff-contact for this stuff. That file is generated from a DB, so I don't think putting it on GitHub is technically an option.

foolip:
@deniak, could you have the things in https://github.com/tobie/specref/blob/master/overwrites/w3c.json fixed in tr.rdf, as well as the things in this PR?

sideshowbarker:
> > @sideshowbarker, is this a problem that could be resolved W3C-side by putting tr.rdf on GitHub? I've sent patches by email, but it's just too much overhead to keep track of.
>
> iirc @deniak is staff-contact for this stuff. That file is generated from a DB, so I don't think putting it on GitHub is technically an option.

Right, it’s a generated file and while I’m not sure @deniak is responsible for the system that generates it, I am sure he can speak from knowledge about what needs to be done to get it fixed/updated.

deniak:
I updated the editor's drafts for Encrypted Media, Touch Events, Web Crypto API, Service Workers 1, UI Events and Vribration  in tr.rdf.
@foolip I'll deal with the ones you mentioned in the email you sent in a bit and will reply when it's done.

foolip:
Thanks @deniak! Even if tr.rdf itself can't be put on GitHub, maybe there could be an issue tracker for things like this? Or tr.rdf is checked in to a repo so that people can suggest changes but need not care about how tr.rdf is generated. The issues could be closed when the published version matches.

Closing this, thanks all!

PR #407, opened 2017-10-09
Add GitHub repo and issue tracker fields

tobie:
/cc @foolip — would that be helpful to some of your current work?

### DIFF
--- a/lib/bibref.js
+++ b/lib/bibref.js
@@ -20,7 +20,8 @@ var PROPS =  [
     "obsoletedBy",
     "versions",
     "versionOf",
-    "deliveredBy"
+    "deliveredBy",
+    "repository"
 ];
 
 var WG_SHORTNAME_SPECIAL_CASES = {
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -76,7 +76,8 @@
         "authors": [
             "Domenic Denicola"
         ],
-        "rawDate": "2015-12-29"
+        "rawDate": "2015-12-29",
+        "repository": "https://github.com/w3ctag/design-principles"
     },
     "APNG": {
         "authors": [
@@ -234,7 +235,8 @@
         "href": "https://github.com/mozilla/id-specs/blob/prod/browserid/index.md",
         "title": "BrowserID",
         "date": "26 February 2013",
-        "publisher": "Mozilla"
+        "publisher": "Mozilla",
+        "repository": "https://github.com/mozilla/id-specs"
     },
     "C14N2-TestCases": {
         "aliasOf": "xml-c14n2-testcases"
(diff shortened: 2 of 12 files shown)


foolip:
Yep, I'm using the repo information. I think it might be a good idea to make it GitHub-agnostic though, there may well be specs maintained using Git outside of GitHub now or in the future.

tobie:
So what: "repository" and "issueTracker", both strings? or make repository an object with `{ "type":  "github", "url": ... }`?

tobie:
> Just strings seems fine, something that can be passed to git clone.

So `https://github.com/[USERNAME]/[REPO].git` or `git@github.com:[USERNAME]/[REPO].git` or what?

> But, actually, for drafts.csswg.org, lots of specs share a repository, so I'm using both a repo URL and a repo path.

How would you structure the JSON for this?

foolip:
Just `https://github.com/[USERNAME]/[REPO]` is fine, the canonical URL that one visits in the browser also works for `git clone`.

As for the path, in day-to-day I have a `specrepo` + `specpath` entries, both strings. However, this has to be very well maintained in specref for it to be a net positive for me, would you generate this information for everything?

tobie:
> However, this has to be very well maintained in specref for it to be a net positive for me, would you generate this information for everything?

Well, looking at what you have for `specpath` in day-to-day, I'm not sure this would be easy to automate. Let's leave it out for now.

tobie:
However, I find myself relying a lot on the "usename/repo" key to make calls to the API, etc. so I wonder if we couldn't make that more of a first class citizen. Not sure how to best name it, or how it would compare on gitlab or elsewhere.

foolip:
Maybe `github:username/repo`?

tabatkins:
Given the URL, it's not hard to (a) tell that this is a github url, and (b) extract the username/pw.  I do this in Bikeshed to autodetect the document's repo by examining `git remote` output, for example.  So, I'm for just having a URL.

`github:username/repo` is no better than `https://github.com/username/repo`; the parsing is approximately as trivial in either, and the latter lets a human immediately go visit the repository, which is good.  If we do end up needing more structured data later, we can add an object for it.

foolip:
Just the URL is fine with me.

dontcallmedom:
In case this helps, I too have an Editors draft URL mapper to github repos in https://github.com/w3c/spec-dashboard/blob/master/fetch-data/group-repos.js#L69

It takes into account a variety of cases where there is no obvious mapping between the URL and the repo.

To model the exact git location of an editors draft, in a [previous project](https://github.com/dontcallmedom/w3c-editors-draft-tracker/blob/master/map-url-to-vcs.json), I had all the needed information with the following properties: a clone URL, a path, and a branch. But clearly, maintaining that info was problematic (although I guess at least a regular existence check could help)

tobie:
Thanks, @dontcallmedom. Copy/pasted it. Happy to integrate it in a cleaner way if you have a better suggestion.

PR #411, opened 2017-10-11
Prefer tools.ietf.org to datatracker.ietf.org for rtcweb documents

dontcallmedom:
Per https://github.com/w3c/webrtc-pc/issues/1629


tobie:
In the future, feel free to merge changes like these as soon as Travis is green.

PR #416, opened 2017-10-25
Manually add WebVR (not yet in /TR/)

foolip:


foolip:
Sending this separately because there's the complication of https://w3c.github.io/webvr/spec/1.1/ also existing. But the latest is what I want, if someone wants 1.1 that could be a separate entry.

tobie:
> if someone wants 1.1 that could be a separate entry.

agreed.

PR #423, opened 2017-11-24
Add MEDIAQUERIES and update MEDIAQ

annevk:
I don't quite understand why this didn't exist or why this is all versioned still, but so be it I guess. We better update these whenever -5 comes out.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -1573,7 +1573,10 @@
         "aliasOf": "mediaont-api-1.0"
     },
     "MEDIAQ": {
-        "aliasOf": "css3-mediaqueries"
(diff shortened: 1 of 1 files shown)


tobie:
Leveled specs are a mess.

annevk:
I guess I can't touch MEDIAQ?

annevk:
I just noticed I made a typo. I wrote MEDIAQEURIES instead of MEDIAQUERIES. Sigh.

tobie:
:D you can either fix it yourself (you'll have to remove it from the json file that lists all of the spec short names so the tests pass), or I'll do it once I've dropped the kids to school.

annevk:
#424.

PR #426, opened 2017-11-29
Update IEEE 754 reference

TimothyGu:


refs/biblio.json
```diff
@@ -1352,9 +1352,22 @@
...
+        "versions": {
+            "1985": {
+                "href": "http://ieeexplore.ieee.org/servlet/opac?punumber=2355",
+                "title": "IEEE Standard for Binary Floating-Point Arithmetic",
+                "publisher": "Institute of Electrical and Electronics Engineers",
+                "rawDate": "1985",
+                "isSuperseded": true
+            },
+            "2008": {
+                "aliasOf": "IEEE-754"
```
tobie: I'm actually surprised this doesn't create an infinite loop of some sort. Still, I feel uncomfortable leaving it around.

Mind changing that instead to:

```json
"2008": {
        "rawDate": "2008-08-29",
         "href": "http://ieeexplore.ieee.org/servlet/opac?punumber=4610933"
}
```
Feel free to squash and merge once you do.

marcoscaceres: @TimothyGu ping

tobie: Turns out @TimothyGu's solution works and mine gets rejected by Specref. The ideal fix here would be to fix the test suite. In the meantime, I'll revert to Timothy's original solution and merge that. Apologies for the the mistakes.

PR #430, opened 2017-12-07
Move SGML-CATALOG from legacy.json to biblio.json

joanmarie:
There is a reference in the ARIA spec.


joanmarie:
@tobie: I thought this change would be sufficient to fix ReSpec from complaining about legacy references. Was there something else I needed to modify in specref?

PR #433, opened 2018-01-31
Add back ED of Selectors

foolip:
Removed in commit c89ab9e1c10db041e7c3c36c02a7a25b606de11e.


foolip:
https://www.w3.org/TR/selectors-3/ doesn't link to it, but it's still true, and without this it looks from the specref data like Selectors is a TR-only spec. Was the cause of this breakage:
https://travis-ci.org/foolip/day-to-day/builds/335059452

tobie:
Yeah, the CSSWG names migration is super painful and creates lots of issues.

PR #436, opened 2018-02-13
Add Interledger references

adrianhopebailie:
This is an interim step to allow us to remove the `localBiblio` from: https://w3c.github.io/webpayments/proposals/interledger/

The long term plan is to produce this file whenever our publishing job runs and make it available at:
https://interledger.org/rfcs/specref.json

When this is complete we will submit a PR with modifications to the update job that pulls the data from this resource.


tobie:
Hey, thanks for making this.

Is there any benefit (on your side) in self-hosting the src file rather than having commit access to this repository and simply updating the biblio.json file?

test/ref-list.json
```diff
@@ -9080,6 +9080,8 @@
     "INSPIRE-MD-20131029",
     "inspire-md-20170302",
     "INSPIRE-MD-20170302",
+    "interledger-protocol",
```
tobie: Fine to keep those in, but for future reference, we have a bot that's going to add them in automatically.

adrianhopebailie: 👍  just wanted to make sure the tests were still passing as expected

adrianhopebailie:
> Is there any benefit (on your side) in self-hosting the src file rather than having commit access to this repository and simply updating the biblio.json file?

Primarily separation of concerns. 

It won't be too difficult for us to produce an `interledger.json` each time we publish and at least this way we avoid a mistake on our side breaking anything on your side.

tobie:
Sorry for not getting back to you earlier, I usually have a much shorter turnaround, but I was at a conference yesterday and then travelling until now.

> It won't be too difficult for us to produce an interledger.json each time we publish […]

Typically specifications tend to add reference at a pace that doesn't make a dedicated file per specification a rationale thing to do in the repository, at least at first. If it turns out that there's a largish (>20) corpus of references a number of related specs are going to rely on, I'd be happy to reconsider my position. In the meantime, though, I'd much rather avoid having to setup the plumbing to pull your data on a regular basis, and instead have you include it in the biblio.json directly.

> […] and at least this way we avoid a mistake on our side breaking anything on your side.

Deployment only happens if the test suite runs properly, so there's absolutely no chance for you to break something accidentally. Of course, with commit access, you can break stuff intentionally, but that's something I'm too worried about at this point. ;)

On the other hand, the plumbing required to fetch external JSON files and pull them in is less resilient to such issues (it'll just crash and email me), so I'd rather limit the number of resources it pulls in for that very reason.

adrianhopebailie:
> Typically specifications tend to add reference at a pace that doesn't make a dedicated file per specification a rationale thing to do in the repository, at least at first.

Sorry, there is a misunderstanding here. The [W3C spec](https://w3c.github.io/webpayments/proposals/interledger/) that references these Interledger RFCs is the first of potentially many (we'd also like to start cross-referencing properly) that references a subset of all of the IL RFCs.

This submission is not intended to be just the set of references for a single spec but rather the beginning of a catalogue of all IL RFCs. We already have 26 with numerous historic drafts for each.

Repo: https://github.com/interledger/rfcs
IL RFC on the website: https://interledger.org/rfcs/0001-interledger-architecture/

This is just my manual submission until I can adjust our publishing scripts to produce the full list.

> the plumbing required to fetch external JSON files and pull them in is less resilient to such issues (it'll just crash and email me), so I'd rather limit the number of resources it pulls in for that very reason.

No problem. I would still prefer to produce an `interledger.json` and push that as part of our automated process if that's okay, rather than try to merge into `biblio.json`.

In summary:
This PR will unblock https://github.com/w3c/webpayments/pull/247 with a small set of IL RFC references that are required by that spec.
I will update our publishing process to produce a much larger `interledger.json` with all 26 IL RFCs including back-versions and this will ultimately replace this manually created file.

tobie:
OK, now I get it. :)

tobie:

Don't hesitate to comment back here if you can't find your changes within 10 minutes; it's a sign something is off and I'll look into it right away. Especially since this is an intermediary step which I haven't really ever tried before, tbh.

Please read-up on how to make [manual changes](https://github.com/tobie/specref#manual-changes), get non-trivial changes reviewed by someone (I'm sure you'll be a good judge of when that's necessary, but typically, I'd like to review the PR where you replace this temporary JSON file with a script to pull in the file you'll host) and feel free to ask if something's unclear.

Thanks again!

PR #446, opened 2018-03-16
Fix ServiceWorker ED URL

### LINKED ISSUE #445: Add current service workers draft
https://w3c.github.io/ServiceWorker/ is not in the database. Ideally [SERVICE-WORKERS] should go there, not to the v1 snapshot.

tobie:
Closes #445.

### DIFF
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -19,5 +19,8 @@
     },
     "encoding": {
         "delete": true
+    },
+    "service-workers-1": {
+        "edDraft": "https://w3c.github.io/ServiceWorker/"
     }
 }


tabatkins:
Yes, Bikeshed distinguishes between "current" and "dated" versions of references; `edDraft` in SpecRef corresponds to a "current" version.  You can request one or the other on an individual biblio-link basis, or specify that you prefer one or the other when both exist, and in the absence of any preference Bikeshed will auto-choose the current or dated version based on whether your `Status` is classified as a current or dated status.

marcoscaceres:
@tabatkins, thanks for the info. I'll take a look at what Bikeshed does, and see I can get ReSpec to follow.

Issue #450, opened 2018-04-04
Reverse lookup on XHR should return WHATWG version, not TR Note


tidoust:
A reverse lookup on the XHR spec URL `https://xhr.spec.whatwg.org/` yields information about the W3C XHR Note, whereas I would expect to get some info about the WHATWG spec:
```
https://api.specref.org/reverse-lookup?urls=https://xhr.spec.whatwg.org/
```

Right now, it returns the same info as:
```
https://api.specref.org/bibrefs?refs=xmlhttprequest
```

... and I would rather expect to get the same info as:
```
https://api.specref.org/bibrefs?refs=xhr
```

I haven't looked at the code, but given that the `https://xhr.spec.whatwg.org/` URL is also the URL of the Editor's Draft of the W3C Note, I would suspect that the code simply returns the first match on the requested URL it finds in the database. Reverse lookup is somewhat fuzzy by definition, but I think it should give priority to specs whose URL matches the requested one, and only return specs whose Editor's Draft URL matches the requested one as a fallback.

tobie:
Oh, that's a tough one. Generally, URLs are supposed to be unique, so something weird must have happened there. Maybe the logic doesn't check for the duplicates between the `edDraft` and `href` properties?

tidoust:
I looked at the code and prepared a PR accordingly. The logic seemed to be _willing to_ check for duplicates, but was not doing it properly in practice...

PR #452, opened 2018-04-09
Give priority to href over edDraft in reverse lookup

tidoust:
Fix for #450.

Code that populated the reverse lookup table seemed to be willing to give priority to `href` over `edDraft`, but it had 2 bugs:
1. it checked for `edDraft` in the properties of the output object not to override a previous entry, but the entry that is inserted is `normalizeUrl(edDraft)`, not `edDraft`
2. it did not add all `href` entries before looking at `edDraft`.

New test that makes sure that the returned ref is always the one that matches `href` if there is one added.

### DIFF
--- a/lib/bibref.js
+++ b/lib/bibref.js
@@ -130,10 +130,15 @@ function prepareReverseLookup(refs) {
             href = normalizeUrl(href);
             output[href] = refs[ref.versionOf || k];
         }
+    });
+    Object.keys(refs).forEach(function(k) {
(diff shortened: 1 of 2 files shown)


tobie:
Good catch! Thanks a bunch. And thanks for the tests!

A new version of Specref updated with these changes should auto-deploy within minutes. You can search [www.specref.org](http://www.specref.org/) to check for it.

Thanks again!

PR #457, opened 2018-04-24
Fixed DCAT-AP and GeoDCAT-AP entries

andrea-perego:
- Fixed version date in DCAT-AP
- Replaced original URLs with the corresponding permanent URIs

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -3265,6 +3265,15 @@
                 "deliveredBy": [
                     "https://joinup.ec.europa.eu/solution/dcat-application-profile-data-portals-europe"
                 ]
+            },
+            "20170224": {
+                "href": "http://data.europa.eu/w21/be3842a8-f87c-4d52-a209-cc21107f20b8",
+                "title": "DCAT Application Profile for data portals in Europe. Version 1.1",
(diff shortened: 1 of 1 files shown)


tobie:
What happened there?

andrea-perego:
I was trying to fix the date used as version key (which I'm quite sure it is not used anywhere), by replacing it with the correct date, but this change didn't pass the validation.

tobie:
Oh, gotcha!

What you want to do in such cases is to add an alias of the incorrect key to the fixed one. Like that you're sure you don't break anything.

andrea-perego:
Thanks, @tobie . I guessed that too late. 

BTW, can aliases be specified at the version level? I mean, like this

````json
"spec": {
  ...
  "versions":{
    "20101010":{
      "aliasOf":"20202020"
    },
    "20202020":{
      ...
    },
  }
}
````

or they should rather be at the top level?

````json
"spec-20101010":{
  "aliasOf":"20202020"
},

"spec": {
  ...
  "versions":{
    "20202020":{
      ...
    },
  }
}
````

tobie:
I think both work.

andrea-perego:
I'll give it a try, then.

PR #460, opened 2018-04-25
Update README to clarify commit rights and review policy

tobie:
/cc @marcoscaceres
 
@andrea-perego, would love your thoughts on whether that gives the right amount of context for you to make the right decisions.


tobie:
> As an aside, would there be an value in putting this into a CONTRIBUTING.md instead?

Yeah, realized there wasn't one right after sending this pull request. Tracking here: #461.

PR #465, opened 2018-05-02
Add aliases for the unleveled version of each css spec

frivoal:
Made to solve https://github.com/w3c/web-platform-tests/issues/10707

Not 100% sure this is the right way of doing it, but let's start with something naive, and make it better if that doesn't cut it.

This is manually generated from the full list of specs in the css-wg's repo, matching the unleveled spec with the same level as what the csswg server does (at the time of writing).

### DIFF
--- a/refs/unleveled-css-alias.json
+++ b/refs/unleveled-css-alias.json
@@ -1,0 +1,215 @@
+{
+    "css-align": {
+        "aliasOf": "css-align-3"
+    },
+    "css-animations": {
+        "aliasOf": "css-animations-1"
+    },
+    "css-backgrounds": {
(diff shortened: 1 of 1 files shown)


tobie:
Unfortunately, it's not as simple and you'll need to go manually through the ones contained in `biblio.json` and `w3c.json` and find duplicates.

The main problem is some of the aliases defined in `w3c.json` might be automatically generated, so even if you remove them, they be added on next update, and break the build.

You're probably better off adding your new aliases directly to the `w3c.json` file and running the test suite to make sure you're not adding duplicates.

tobie:
Feel free to reopen this if you want to implement the requested changes.

PR #470, opened 2018-06-20
Update WG21 references, include WG21 in auto-update scripts

### LINKED ISSUE #469: WG21 missing entries
WG21 seems to be missing entries, for example:

screen shot 2018-06-20 at 2 19 25 pm

Looking at wg21.link/index.json there are r0 through r6 versions of this paper, but specref doesn't know about all of them.

/cc @m-ou-se @jyasskin

tobie:
Closes #469.


jfbastien:
Confirmed that it now works (and bikeshed is happy again). Thank!

tobie:
Is your version of Bikeshed up to date?

jfbastien:
Yeah I ran `bikeshed update` just before filing the issue. After your fix I did `bikeshed update` again and it found the paper I was trying to refer to.

tobie:
That updates the data, not the code, for which you have to use git.

jfbastien:
Right but the data was what I needed, no? Unless bikeshed started using specref differently?

tobie:
Yeah. It pulls WG21 directly.

m-ou-se:
Note that wg21.link also publishes its index in specref format: https://wg21.link/specref.json
Might be useful.

PR #472, opened 2018-07-06
Move "Reporting" from WICG to W3C

paulmeyer90:


paulmeyer90:
@tobie, please review.

refs/biblio.json
```diff
@@ -2053,6 +2053,13 @@
         "date": "2008",
         "isoNumber": "ISO/IEC 19757-2:2008"
     },
+    "REPORTING": {
+        "href": "https://w3c.github.io/reporting/",
+        "title": "Reporting API",
+        "status": "Editor's Draft",
```
tobie: Use ED for status.

clelland:
@tobie, the spec is kind of in limbo at the moment -- it was moved from [WICG/reporitng](https://github.com/WICG/reporting) to [w3c/reporting](https://github.com/w3c/reporting) a week ago, and that seems to have stopped things like bikeshed from being able to reference it. I figured that a change to the specref database like this would be the way to get bikeshed working again, and that pointing to the ED would still allow the automated processes to pick up the FPWD from tr.rdf if and when that becomes available.

If there's a better process for this, if this isn't the right way to patch this up, we're more than happy to learn and do it right.

tobie:
There’s absolutely no reason that this reference should stop working in Bikeshed. How are you referencing it?

clelland:
(I'll admit that it's possible that this has nothing to do with the problem that started me on this. I probably jumped too quickly to assume causality -- I believed that bikeshed-data used specref to find the source of specs to crawl)

I've tried referencing exported features from the Reporting spec like this (with a freshly-updated bikeshed):

```
 <a>endpoint group</a>
 <a>report</a>
 <a>failover class</a>
{{endpoint/priority}}
 <a>ReportTo</a>
```

All give me link errors. None of the terms appear in `/spec-data/`. Looking for the source, I came to specref.org, and found that the spec is listed in the old (wrong) location --  https://wicg.github.io/reporting doesn't host anything anymore. 

(Using `[[REPORTING]]` works, but references the old location.)

tobie:
Oh. I don’t know what @tabatkins uses for x-refs. (Is it manual or automated?)

Either way, you should fix this PR by adding the references in the w3c.json file rather than in this one.

tobie:
Closing this for now. Please feel free to repoen if you want to fix the changes.

PR #479, opened 2018-10-17
Update "xpath" (XPath 1.0) URL

sideshowbarker:
This change makes https://www.w3.org/TR/xpath-10/ the URL XPath 1.0 (key: "xpath"), rather than https://www.w3.org/TR/xpath (which now redirects to https://www.w3.org/TR/xpath/all/ and has become an index/overview page rather than being the actual XPath 1.0 spec).


tobie:
This is probably the same issue as #480.

sideshowbarker:
Amended and force-pushed 3d6db69 to put the patch where it belongs — but had the same problem as described in https://github.com/tobie/specref/pull/480#issuecomment-430510492 as far as the update script not actually changing anything.

PR #480, opened 2018-10-17
Update "webvr" URLs

sideshowbarker:
This change makes https://immersive-web.github.io/webvr/spec/1.1/ the URL for the WebVR spec (key: "webvr"), rather than the URL https://w3c.github.io/webvr/spec/latest/ (which now redirects to a 404).  The change also updates the "repository" URL to https://github.com/immersive-web/webxr (which is where the old https://github.com/w3c/webvr now redirects to).

PR #481, opened 2018-10-17
Follow-up to #480 and #479

tobie:
/cc @sideshowbarker

### DIFF
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -23,10 +23,6 @@
     "service-workers-1": {
         "edDraft": { "replaceWith": "https://w3c.github.io/ServiceWorker/" }
     },
-    "webvr": {
-        "href": { "replaceWith": "https://immersive-web.github.io/webvr/spec/1.1/" },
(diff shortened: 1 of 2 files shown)


tobie:
Thanks for the heads-up @deniak.

overwrites/w3c.json
```diff
@@ -23,10 +23,6 @@
     "service-workers-1": {
         "edDraft": { "replaceWith": "https://w3c.github.io/ServiceWorker/" }
     },
-    "webvr": {
-        "href": { "replaceWith": "https://immersive-web.github.io/webvr/spec/1.1/" },
-        "repository": { "replaceWith": "https://github.com/immersive-web/webxr" }
-    },
     "xpath": {
```
sideshowbarker: The `"xpath"` key and value now also need to be removed from `overwrites/w3c.json`, no?

tobie: No. Because that one actually is in W3C's database and has to be overwritten, if not, it'll just revert to its previous state within the next hour.

PR #490, opened 2018-12-11
Updated ETSI TS 102366 to latest version

wschidol:
- changed title field to include ETSI TS number and exclude version
- added deliveredBy


wschidol:
> I assume `deliveredBy` is what's causing issues. I'd drop it altogether if I was you.

Thanks for the hint -- indeed, this makes the build work. So if I get the checkin rules correctly, I now wait an "appropriate amount" of time to give others a chance to review, then check in. What would you consider "appropriate" in this case? 1 day maybe since it is a more-or-less trivial change?

PR #492, opened 2018-12-11
Added UN/CEFACT

JanWerder:
This PR adds the source for the United Nations Centre for Trade Facilitation and Electronic Business Core Components Data Type Catalogue.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -4645,6 +4645,16 @@
         "status": "Normative",
         "rawDate": "2015-03-01",
         "title": "OMG Unified Modeling Language"
+    },        
+    "UN/CEFACT": {
+        "authors": [
+            "United Nations Economic Commission for Europe"
+        ],
+        "href": "https://www.unece.org/cefact/codesfortrade/ccts_datatypecatalogue.html",
+        "publisher": "United Nations Economic Commission for Europe",        
+        "status": "Normative",
+        "rawDate": "2011-09-17",
+        "title": "United Nations Centre for Trade Facilitation and Electronic Business Core Components Data Type Catalogue"
     },
     "vann": {
         "authors": [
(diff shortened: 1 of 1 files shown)


refs/biblio.json
```diff
@@ -4645,6 +4645,16 @@
         "status": "Normative",
         "rawDate": "2015-03-01",
         "title": "OMG Unified Modeling Language"
+    },        
+    "UN/CEFACT": {
```
marcoscaceres: Im not sure we have used “/“ as a divider before... or have we?

JanWerder: It doesn't seem so. I could replace it with a dash if that fits the style of the other sources better.

marcoscaceres: Yeah, might be better.

tobie: Yeah. That’s probably why the test suite is balking.

refs/biblio.json
```diff
@@ -4645,6 +4645,16 @@
...
+        "authors": [
+            "United Nations Economic Commission for Europe"
+        ],
+        "href": "https://www.unece.org/cefact/codesfortrade/ccts_datatypecatalogue.html",
```
marcoscaceres: Looking at this URL, it seems to link to 3 pdfs. Could you help us understand how you intend to cite this work in practice?

JanWerder: I thought that whenever this source is linked the referenced version should come afterwards. But I could change the URL to the pdf with the most recent version.

refs/biblio.json
JanWerder: Yep, it passes now without the slash

refs/biblio.json
marcoscaceres: How about keeping this one as “latest”, but the adding UN-CEFACT-V1, UN-CEFACT-V2, UN-CEFACT-V3? Or would that be overkill? Alternatively, take a look at how other specs that have versions deal with this. Just a suggestion, as you might know best how these documents are to be cited.

PR #508, opened 2019-02-13
Add HLS Timed Metadata

chrisn:
I have a couple more entries to add. Thanks for your help!


chrisn:
Thanks, and fixed. Is this OK to merge?

tobie:
If it’s green and reasonable you can always go ahead.

PR #513, opened 2019-03-01
JAXRS20

namedgraph:


refs/biblio.json
```diff
@@ -1685,6 +1685,14 @@
         "title": "Java Accessibility API",
         "publisher": "Oracle"
     },
+    "JAXRS20": {
```
marcoscaceres: Probably want to keep the same key.

```suggestion
    "JAX-RS": {
```

You can also add an alias for JAXRS20.

namedgraph: Done. The latest version is 2.1 though - does that matter?

marcoscaceres: Depends. If 2.1 includes all of 2.0, and 2.1 is "stable" as per whatever that community considers stable, then it would be best to update this to 2.1. 

If 2.1 is incompatible or somehow conflicting with 2.0... or folks are legally required to cite 2.0, for instance, then a new entry might be needed.

namedgraph: Done

PR #525, opened 2019-04-12
Define all ISO/IEC JTC1 SC29 documents

sandersaares:
Imported from ISO website for JTC1 SC29 ("Coding of audio, picture, multimedia and hypermedia information").

Duplicates already in biblio.json aliased.

JSON generated using https://github.com/sandersaares/ScrapeIsoDocuments

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -1609,13 +1609,6 @@
         "title": "Integrated Services Digital Broadcasting",
         "href": "http://dibeg.org/"
     },
-    "ISO14496-30": {
-        "title": "Information technology — Coding of audio-visual objects — Part 30: Timed text and other visual overlays in ISO base media file format",
-        "href": "http://www.iso.org/iso/home/store/catalogue_tc/catalogue_detail.htm?csnumber=63107",
-        "date": "11 March 2014",
-        "publisher": "ISO/IEC",
-        "isoNumber": "ISO 14496-30:2014"
-    },
     "ISO3166": {
         "aliasOf": "ISO3166-1"
     },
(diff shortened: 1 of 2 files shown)


tobie:
Thank you for this patch.

Specref has an auto-update mechanism by which it polls data sources hourly to stay up to date. By the look of this data, this collection was automated. Would you be willing to include a script to do so here?

See for example:

* https://github.com/tobie/specref/blob/master/scripts/fetch-helpers/wg21.js (using the fetch-helper), or
* https://github.com/tobie/specref/blob/master/scripts/w3c.js (for a more involved one parsing an RDF file).

sandersaares:
I have the scraper in the form of a C# app - looks like the scripts in this repo are JavaScript, so there is a bit of a mismatch. Hmm, tell you what, I will just schedule a run of my C# app to do the scrape every 24 hours and publish the results online. Then you can just import the latest copy with whatever automation you use.

Would that work for automatic updates? I went ahead and set up the automated publishing; the file is available at https://dashifspecref.blob.core.windows.net/iso/iso_jtc1_sc29.json?st=2019-04-01T00%3A00%3A00Z&se=2039-04-01T00%3A00%3A00Z&sp=rl&sv=2018-03-28&sr=b&sig=Q1%2F6yLw90v0VCYKPhuSw89qLXS8tPZWU%2BQSvYj%2B7NmY%3D and will be updated around 01:00 UTC daily.

andrea-perego:
Thanks for this, @sandersaares .

I noticed a couple of things in the resulting JSON that you may consider fixing:
- The `href` URLs of the relevant specs always include a query string - e.g., `https://www.iso.org/standard/18902.html?browse=tc` . AFAIK, the corresponding canonical URL (`https://www.iso.org/standard/18902.html`) is not supposed to have itt
- The entries are missing the date/year of publication

sandersaares:
Both fixed now, will be included in next update. URL fix also brought to surface some missing aliases (#530). Will submit new PR to include rawDate and also obsoletedBy

sandersaares:
After fixing URLs the update script did something strange and moved "MP3" from biblio.json to iso_jtc1_sc29.json and aliased it to nonexisting `iso_jtc1_sc29-iso11172-3`

I assume my manual aliasing in #530 will avoid this incorrect aliasing but @tobie perhaps you should take a look - I do not quite understand the update script logic enough to fix it myself here.

PR #526, opened 2019-04-13
Automate iso jtc

tobie:
/cc @sandersaares


sandersaares:
@tobie I have added "obsoletedBy" values to the output of my scrape script. Not sure what needs to be adjusted in the specref scripts to make use of it, so I leave that part up to you if you will.

I also notice that the ISO website where I get the entries from actually seems to be deleting some obsolete references. My code just converts the website to JSON, so does not do anything to preserve historical entries. Will this be a problem?

Issue #528, opened 2019-04-17
"npm test" rejects ' ' in URL but "fetch-refs" puts it there


sandersaares:
Example URL: `https://www.scte.org/SCTEDocs/Standards/ANSI_SCTE 01 2015.pdf`

I try to add references like this but `npm test` complains that this is not a valid URL. Because of the space characters. Okay, so I fix it to use %20 - works for me. But then if I try to add an update script, fetch-refs adds the space back in original form.

I think "npm test" should accept space in URLs.

tobie:
That seems like something that needs to be fixed in the module that fetch-refs loads.

tobie:
I think my suggestion in #529 fixes that? Feel free to reopen otherwise.

sandersaares:
I agree, it fixes the issue or at least works around it for this specific case. I am not sure if rejecting spaces is the best path in terms of compatibility - many things seem to consider spaces fine without escaping - but I suppose that is a secondary topic.

PR #529, opened 2019-04-17
Add SCTE standards

sandersaares:
This is the full standards catalog of [SCTE](https://www.scte.org/) (Society of Cable Telecommunications Engineers).

Scraped from website using https://github.com/sandersaares/ScrapeScteDocuments and published every 24 hours. Update script included based on ISO example.

**Blocked** by https://github.com/tobie/specref/issues/528


scripts/fetch-helpers/scte.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function(id, ref) {
+    return {
+        href: ref.href,
```
tobie:
```suggestion
        href: encodeURI(ref.href),
```

Issue #532, opened 2019-04-19
Stale data?


marcoscaceres:
Seems the WICG and W3C data might be a little stale?... looking at the commit logs, W3C was only updated 6 days ago and WICG was 20 days ago (both datasets have had updates since then). @tobie, maybe an update script is stuck somewhere?

tobie:
Yes. Will take a look shortly. Sorry for the issue.

tobie:
Fixed. A typo in an added automated script.

marcoscaceres:
Yay! Thanks so much!

PR #536, opened 2019-04-23
Move feature policy to W3C

### LINKED ISSUE #533: Move Feature Policy from WICG to W3C
It is on /TR now: https://www.w3.org/TR/feature-policy-1/

tobie:
Closes #533.

### DIFF
--- a/refs/w3c.json
+++ b/refs/w3c.json
@@ -31595,6 +31595,9 @@
         "rawDate": "1997-10-28",
         "title": "Testbed"
     },
+    "feature-policy": {
+      "aliasOf": "feature-policy-1"
+    },
     "feature-policy-1": {
         "authors": [
             "Ian Clelland"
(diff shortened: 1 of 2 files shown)


marcoscaceres:
Note that I updated the WICG file downstream... #532 ?

tobie:
So… the problem here is a bit more complex. There are safeguards to keep the data consistent against upstream changes. The problem is, until the spec surfaces in /TR (i.e. until the first public WD), allowing it to be removed from the WICG refs file would break a bunch of specs referencing it.

marcoscaceres:
Agree and understood. I only made the change once it was on TR:
https://www.w3.org/TR/feature-policy/

tobie:
Yes... except for some unknown reasons, the /TR RDF API exposes the new FEATURE-POLICY-1 URL, but not the FEATURE-POLICY -> FEATURE-POLICY-1 redirect (it usually does).

marcoscaceres:
I'm just going to go and smash my head against the wall on your behalf for a while.

tobie:
Give it a good 90 minutes.

PR #545, opened 2019-06-13
Added reference to the OpenAPI specification

vcharpenay:


refs/biblio.json
```diff
@@ -2124,6 +2124,26 @@
...
+    "OPENAPIS": {
+        "href": "https://www.openapis.org/",
+        "title": "OpenAPI Specification",
+        "publisher": "OpenAPI Initiative",
+        "versions": {
+            "3.0.2": {
+                "href": "http://spec.openapis.org/oas/v3.0.2",
+                "title": "OpenAPI Specification - version 3.0.2",
+                "rawDate": "2018-10-08",
+                "authors": [
```
tobie: You want to share this data (notably authors) with the main entry, too.

vcharpenay: thanks for the feedback! See my edits.

Author's list is different for every version, I hope the way I defined them is ok (all authors merged into a single array). Regarding dates, since it is a living standard, I do not know if it appropriate to add the date of a specific version.

andrea-perego:
Thanks for adding this entry, @vcharpenay .

We are actually referencing OpenAPI in [the new version of the DCAT vocabulary](https://www.w3.org/TR/vocab-dcat-2/#bib-openapi), so having it in SpecRef is very handy.

Just a small comment:

Would it be possible to change the entry key from "OPENAPIS" to "OPENAPI"? I guess the trailing "S" stands for "Specification", but it is more a common practice just to use the name / acronym of the standard / technology, with "specification", "vocabulary", "schema", etc. Moreover, the trailing "S" may make it difficult to find the entry for people looking for it.

vcharpenay:
Hi @andrea-perego,

I actually took the host name of OAI's web server: `openapis.org`. I do not know myself what it stands for exactly but I guess it is how the OAI wants to advertise their content. Regarding indexing, you will find the entry even if [you just type `openapi`](https://www.specref.org/?q=openapi) in SpecRef's search tool.

Now that the PR is closed, you can maybe add yourself an alias to the `OPENAPIS` entry?

andrea-perego:
Thanks, @vcharpenay . I'll can take care of the alias.

PR #556, opened 2019-08-01
Add an overwrite file for WICG

tobie:
/cc @tidoust


tobie:
I _think_ it should be temporary and that should have fixed it?

marcoscaceres:
it should have.

Issue #560, opened 2019-10-13
tc39.github.io redirects to tc39.es/ 


contrepoint:
Passed the [Web App Manifest url](https://www.w3.org/TR/appmanifest/) into W3C's [link checker](https://validator.w3.org/checklink?uri=https://www.w3.org/TR/appmanifest/&recursive=on) and found this redirect for ECMA-402. 

I will update the link in a PR

```  Line: 6529 https://tc39.github.io/ecma402/ redirected to https://tc39.es/ecma402/ Status: 301 -> 200 OK This is a permanent redirect. The link should be updated. ```

ETA: updated issue title to reflect @tobie's observation below. thank you!

tobie:
Sounds like all GitHub hosted TC-39 specs should be affected by this, no? Think you could do a broader PR and either check the links yourself or get someone from TC39 to approve it?

tobie:
Also, please note that the deployment pipeline is currently down because of a broken update and your changes won’t be deployed until Friday.

contrepoint:
@tobie yes - that is true, thanks for pointing it out! I'll make another PR tomorrow to update the links for the other tc-39 specs.

tobie:
Awesome, thanks!

PR #562, opened 2019-10-14
update tc39 links

contrepoint:
closes [issue 560](https://github.com/tobie/specref/issues/560)

@tobie -
* I searched the entire repo for `tc39` and there was only one remaining link. I updated it to point to `tc39.es` and checked to make sure that `https://tc39.github.io/ecma262/` redirects to `https://tc39.es/ecma262/`
* Thanks in advance for your review!

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -1109,7 +1109,7 @@
     },
     "ECMASCRIPT": {
         "title": "ECMAScript Language Specification",
-        "href": "https://tc39.github.io/ecma262/",
+        "href": "https://tc39.es/ecma262/",
         "publisher": "Ecma International",
         "versions": {
             "51": {


marcoscaceres:
Any hits for “ecma”? It might include a few more.

contrepoint:
@marcoscaceres

there are hits, but none use that domain and there are no redirects.
1. `ECMA-404` points to a [pdf](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)
2. Other versions of `ECMA-262` point to [ecma-international.org](http://www.ecma-international.org/). For example, version 10.0 points to http://www.ecma-international.org/ecma-262/10.0/index.html. `10.0` is the version number and changes depending on the ECMA version being referenced.
3. The rest are aliases. For example, `ECMA-262` is the alias of `ECMASCRIPT`

marcoscaceres:
Ok, cool. Thanks for checking!

PR #564, opened 2019-10-29
Add alias for all CSS specs

### LINKED ISSUE #563: Clarification on CSS Cascade
Hi @tabatkins ... we have an interesting issue. Terms like `[=used value=]` are defined in both "css-cascade-3" and "css-cascade-4", but under the short name "css-cascade". 

When citing, ideally authors would have the option to just cite "css-cascade", meaning "the latest" or "highest level". However, ""css-cascade" is not aliased in Specref.... we are unsure if:

1. we should explicitly alias "css-cascade" to "css-cascade-4" in biblio.js? 
2. in xref, just always assume that "css-cascade" means  "css-cascade-4" (or the one with the highest level)

Right now, the we are overcoming this problem by using a localBiblio:

```
        localBiblio: {
          "css-cascade": {
            aliasOf: "css-cascade-4"
          }
        }
```

But that doesn't seem great.

marcoscaceres:
closes #563


marcoscaceres:
@tobie, sorry, it's been a while since I worked on SpecRef... I was unsure if I should update "refs/w3c.json" or "biblio.json"? I guess "w3c.json" is right? 

Also, I'm unsure what to do about the failing tests. Could you guide me a little bit?

refs/w3c.json
```diff
@@ -21662,6 +21713,9 @@
         "edDraft": "https://drafts.csswg.org/css-gcpm/",
         "repository": "https://github.com/w3c/csswg-drafts"
     },
+    "css-grid": {
+        "aliasOf": "css-grid-2"
```
tabatkins:
```suggestion
        "aliasOf": "css-grid-1"
```

(grid-2 is still a delta spec for now)

marcoscaceres: Ok, I added a "overrides" map to the script to handle these kinds of cases.

tobie:
This is a great.

The reasons the tests are failing is that some of these aliases have already been added manually, probably both in `w3c.json` and `biblio.json`.  Removing them manually once should fix it, as long as all of them are aliases (and none actually contain data).

Is there anything preventing us from adding the script to `update-all` so that it runs with automatic updates?

marcoscaceres:
> Is there anything preventing us from adding the script to update-all so that it runs with automatic updates?

No, that was going to be my next question :)

tobie:
Awesome. Let’s do that. I have a few suggestions on the script itself that I’ll get to right away (for consistency with the rest of this mess🙃) and then we should be good to go.

tobie:
Whereby _rest of this mess_ I mean the whole code base, not your pull request.

marcoscaceres:
> Whereby rest of this mess I mean the whole code base, not your pull request.

It's all beautiful soup :) I also slapped mine together rather quickly, so appreciate the look over.

scripts/update-all
```diff
@@ -52,7 +52,9 @@ function run()
     node "./scripts/run.js" "./github" "w3c.json" &&
     node "./scripts/run.js" "./github" "wicg.json" &&
     node "./scripts/run.js" "./github" "biblio.json" &&
-    node "./scripts/run.js" "./github" "whatwg.json";
+    node "./scripts/run.js" "./github" "whatwg.json" &&
+    node "./scripts/scripts/css-alias.js" &&
+    node "./scripts/list-refs.js"; # regenerate, as CSS aliases may have changed
```
tobie: Why are you regenerating that list here? This seems like a pretty big smell.

refs/wicg.json
```diff
@@ -47,14 +47,6 @@
         "source": "https://wicg.github.io/admin/biblio.json",
         "repository": "https://github.com/wicg/cors-rfc1918"
     },
-    "CSS-ANIMATION-WORKLET": {
```
tobie: Where did that one go?

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * Script creates alias for CSS specs
+ * see https://github.com/tobie/specref/issues/563
+ */
+const { promises: fs } = require("fs");
+const w3cSpecsPath = require("path").resolve(__dirname, "../refs/w3c.json");
```
tobie: var current = helper.readBiblio(FILE);
```suggestion
const helper = require('./helper');
const FILENAME = "w3c.json";
const obj = helper.readBiblio(FILENAME);
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
...
+
+// Some are delta specs, or not stable enough to
+// be used as the canonical ones.
+const overrides = new Map([["css-grid", "css-grid-1"]]);
+
+(async () => {
```
tobie:
```suggestion
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
...
+    .sort()
+    .reduce((sortedObj, key) => {
+      sortedObj[key] = obj[key];
+      return sortedObj;
+    }, {});
+
+  // Write the file back out
+  const json = JSON.stringify(sortedObj, null, 4) + "\n";
+  await fs.writeFile(w3cSpecsPath, json, "utf-8");
+})();
```
tobie:
```suggestion
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
...
+  // Insert the aliasOf entries in the right order
+  const sortedObj = Object.keys(obj)
+    .sort()
+    .reduce((sortedObj, key) => {
+      sortedObj[key] = obj[key];
+      return sortedObj;
+    }, {});
+
+  // Write the file back out
+  const json = JSON.stringify(sortedObj, null, 4) + "\n";
```
tobie:
```suggestion
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+const { promises: fs } = require("fs");
+const w3cSpecsPath = require("path").resolve(__dirname, "../refs/w3c.json");
+
+// Some are delta specs, or not stable enough to
+// be used as the canonical ones.
+const overrides = new Map([["css-grid", "css-grid-1"]]);
+
+(async () => {
+  const data = await fs.readFile(w3cSpecsPath);
```
tobie:
```suggestion
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
...
+  const obj = JSON.parse(data);
```
tobie:
```suggestion
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
...
+  const sortedObj = Object.keys(obj)
+    .sort()
+    .reduce((sortedObj, key) => {
+      sortedObj[key] = obj[key];
+      return sortedObj;
+    }, {});
+
+  // Write the file back out
+  const json = JSON.stringify(sortedObj, null, 4) + "\n";
+  await fs.writeFile(w3cSpecsPath, json, "utf-8");
```
tobie:
```suggestion
helper.writeBiblio(FILENAME, sortedObj);
```

scripts/css-alias.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * Script creates alias for CSS specs
+ * see https://github.com/tobie/specref/issues/563
+ */
+const { promises: fs } = require("fs");
```
tobie:
```suggestion
```

test/ref-list.json
```diff
@@ -2685,32 +3293,12 @@
...
-    "CSS-VALUES-20150611",
-    "css-values-20160929",
-    "CSS-VALUES-20160929",
     "css-values-20180814",
     "CSS-VALUES-20180814",
+    "css-values-20181010",
+    "CSS-VALUES-20181010",
     "css-values-20190131",
     "CSS-VALUES-20190131",
-    "css-values-20190606",
```
tobie: The whole point of the ref-list file is to make sure entries don't suddenly disappear. That contract is no longer fulfilled if you remove them from the ref-list. You have to automate the aliasing instead.

scripts/update-all
tobie: Yeah, you really can't do that. You're basically by-passing the whole check to verify that you're not removing entries by running this at this point.

Instead you need to figure out what entries are getting removed and add them automatically.

marcoscaceres: If, for example: 

 * today "css-foo" references "css-foo-1", 
 * tomorrow  "css-foo" becomes "css-foo-2"

Now all the "css-foo-YYYYMMDD" references need to be updated... this is where is gets complicated. Perhaps "date citing" an an alias shouldn't be allowed. Or, we need to figure out how to retain these old references as more aliases.

refs/wicg.json
marcoscaceres: The Houdini folks took it. It's in the w3c.json file now.

scripts/update-all
tobie: Absolutely. I now remember why trying to fix this index issue in the past had made me sad.

The correct way of solving the problem is stacking everything onto "foo" including version numbers ("foo-1"), rec track status ("foo-1-CR"), and dated releases ("foo-YYYYMMDD") like so:

```json
{
    "css-grid": {
        "versions": {
            "1": {...},
            "2": {...},
            "20110407": {...},
            "20120322": {...},
            "20121106": {...},
            "20130402": {...},
            "20130910": {...},
            "20140123": {...},
            "20140513": {...},
            "20150317": {...},
            "20150806": {...},
            "20150917": {...},
            "20160519": {...},
            "20160929": {...},
            "20170209": {...},
            "20170509": {...},
            "20171214": {...},
            "20180206": {...},
            "20180427": {...},
            "20180628": {...},
            "20180804": {...}
        }
    }
}
```

Compared to what we have which is more:

```json
{
    "css-grid-1": {
        "versions": {
            "20110407": {...},
            "20120322": {...},
            "20121106": {...},
            "20130402": {...},
            "20130910": {...},
            "20140123": {...},
            "20140513": {...},
            "20150317": {...},
            "20150806": {...},
            "20150917": {...},
            "20160519": {...},
            "20160929": {...},
            "20170209": {...},
            "20170509": {...},
            "20171214": {...}
        }
    },
    "css-grid-2": {
        "versions": {
            "20180206": {...},
            "20180427": {...},
            "20180628": {...},
            "20180804": {...}
        },
    }
}
```

This is actually hard to do given the existing data set.

refs/wicg.json
tobie: So you're telling me it disappeared, right? 🐇✨🎩

scripts/update-all
tobie: Well, I guess even that has some limitations if there are two concurrent versions releasing new drafts on the same day. :-/

tobie: Maybe your best bet is to run this manually from time to time for now?

marcoscaceres: 😭... indeed. ok... argh. Ok, it's doable. Need to find all "css-grid-*" can then clone all `versions` in. Will try tomorrow.

tobie: It might be worth merging this minus the changes to update-all, and really giving the data model some thought before committing to something else, there.

tabatkins: Yeah, having a dated reference to the unleveled ref doesn't make sense; either it only resolves to a particular version's dates (which then become invalid when a new version comes out and the unleveled ref moves) or it resolves to all of them (and then you can't disambiguate when two versions of a spec come out on the same day).  The latter issue is *definitely* possible, especially now that we have such easier publishing via Echidna; I wouldn't be surprised if it had already happened, actually.

marcoscaceres: > Yeah, having a dated reference to the unleveled ref doesn't make sense; 

Agree.

> The latter issue is definitely possible, especially now that we have such easier publishing via Echidna; I wouldn't be surprised if it had already happened, actually.

I has... I spotted one already.

tobie: Alright, so we'll have to maintain the existing ones for backwards compat and start anew with the rest. Maybe worth looking into the JSON API while we're at it?

test/ref-list.json
tobie: Still want that one fixed ;)

marcoscaceres: yep, on it!

marcoscaceres:
I'm struggling a bit :( The changes to "test/ref-list.json" would add unwanted refs keyed on version (which we don't want, based on the discussion here). 

I'll revert the changes to "test/ref-list.json", but I think we still need to figure out how to change some assumptions we've made with referencing the `aliasOf` things that are going to break going forward. 

I don't know yet how to cope with that... we might be in breaking change situation - or we might need a way of hard coding those legacy `aliasOf` generated references in "test/ref-list.json".

test/ref-list.json
marcoscaceres: switched back "css-values-3".

marcoscaceres:
Ok... this gets us back into a "good state". However, running "scripts/list-refs.js" again now will  generate the references we don't want anymore. 

Let's see if we can figure out a good path forward.

ExE-Boss:
It&nbsp;should&nbsp;probably now&nbsp;be&nbsp;possible to&nbsp;implement&nbsp;this using&nbsp;the&nbsp;framework&nbsp;introduced in&nbsp;<https://github.com/tobie/specref/pull/573>.

marcoscaceres:
@ExE-Boss, excellent! Should this be closed or could you provide me with some guidance for next steps?

sidvishnoi:
Any updates on this?
Do we need to manually add all entries to `overwrites/w3c.json`? Can we safely redirect the output of this script to `overwrites/w3c.json` to add `createAlias`entries?

marcoscaceres:
@tobie, @ExE-Boss, was waiting on next steps on this... @sidvishnoi is willing to take it over, but we need some guidance 🙏

tobie:
My feeling is that to move forward with something like this will require a lot of refactoring.

The schema we have doesn’t really handle the different ways the different groups think about versioning. I don’t know that there’s a model that would work across the board, but maybe there are a couple?

Once we have a schema in place we‘ll also need to have have a migration story for the existing data and for the generated references.

Do we keep the latter around or do we deprecate them?

We’ll also probably need to rethink how the generated files are kept separated from one another. That has cause more problems than anticipated. Similarly, whether these should contain manual edits or not should be clarified, so should how overwriting works and why, etc.

Maybe there’s a shortcut here, but I’ve attempted it twice and failed both times.

tobie:
In particular, you want to look at this comment and the following thread: https://github.com/tobie/specref/pull/564#discussion_r340532180

tobie:
Should we close this and reopen something else once we have clarity as to how to move forward?

sidvishnoi:
Agree. Lets close this for now.

Issue #571, opened 2019-12-02
HR-TIME should be an alias for the latest HR-TIME


domenic:
https://www.specref.org/?q=HR-TIME shows `[HR-TIME]` as obsolete, but the DOM Standard needs to be able to refer to the latest HR-TIME ED at https://w3c.github.io/hr-time/ in some way. I would assume `[HR-TIME]` could be that reference, with e.g. `[HR-TIME-1]` being the obsolete one.

tobie:
Heh. That's yet again another one of the same issue of W3C not being consistent with their versioning scheme. You might be able to fix that using overwrites. @marcoscaceres attempted a solution in https://github.com/tobie/specref/pull/564 (although for CSS specs only). But it came with its own set of issues.

The only real solution moving forward would be to enforce a consistent data model in Specref and force W3C specs into it. I don't think anyone really wants to do the work for this, though.

domenic:
Do you have advice for what we should do in the DOM Standard in the meantime? I'd guess, switch to using `[HR-TIME-2]`, and update it every time a new edition comes out?

I guess part of the weirdness is that Bikeshed is complaining at us about using "obsolete" refs, but at the same time using Bikeshed's "prefer latest specs" option means that the obsolete URL never shows up, only the ED URL. /cc @tabatkins

tobie:
> Do you have advice for what we should do in the DOM Standard in the meantime? I'd guess, switch to using [HR-TIME-2], and update it every time a new edition comes out?

Yeah. This sucks. I agree.

domenic:
It's frustrating that when I try that, Bikeshed lists the bibliography entry as "High Resolution Time Level 2. URL: https://w3c.github.io/hr-time/", but when I click on that URL, the title is just "High Resolution Time". More of a Bikeshed issue though, I guess...

domenic:
I guess what's really missing here is an entry in specref that is "Editor's Draft only", which Bikeshed could use when in the `Default Ref Status: current` mode. That would have a title derived from the ED, instead of the TR counterpart.

The problem is now specref is keying off of TR documents, so when a spec author wants to reference an ED, they have to pick a random corresponding TR document and tell Bikeshed to use that specref entry. Bikeshed then dutifully uses that specref entry's title field ("High Resolution Time level 2"), plus its ED URL field (https://w3c.github.io/hr-time/), which causes a mismatch when the ED title and TR title are different.

tobie:
Those are good points.

[...]

PR #572, opened 2019-12-04
fix(refs/w3c): `hr‑time` is an alias for `hr‑time‑2`

ExE-Boss:
See&nbsp;issue&nbsp;<https://github.com/tobie/specref/issues/571>, specifically&nbsp;<https://github.com/tobie/specref/issues/571#issuecomment-561673316>.

---

review?(@tobie)


tobie:
I *think* this would get overwritten by the bot.

scripts/helper.js
```diff
@@ -99,6 +99,16 @@ function tryOverwrite(f) {
...
+            } else if (ow.renameTo) {
+                references[ow.renameTo] = ref;
+                delete references[k];
+                console.log("Renamed", k, "to", ow.renameTo);
+
+                if (ow.aliasOf && ow.aliasOf.replaceWith) {
+                    console.log("Adding", k, "...");
+                    console.log("   ", "aliasOf:", undefined, "->", JSON.stringify(ow.aliasOf.replaceWith));
+                    references[k] = { aliasOf: ow.aliasOf.replaceWith };
+                }
```
tobie: Could this be more succinctly and clearly expressed with something like `switchWith` where you'd write:

```json
{
 "hr-time": {
        "switchWith": "hr-time-1",
    },
    "hr-time-1": {
        "href": { "replaceWith": "https://www.w3.org/TR/hr-time-1/" }
    }
}
```

It seems this is an common enough need that we should cater to it specifically.

refs/w3c.json
```diff
@@ -36597,6 +36600,21 @@
...
+    },
+    "hr-time-20121023": {
+        "aliasOf": "hr-time-1-20121023"
+    },
+    "hr-time-20121217": {
+        "aliasOf": "hr-time-1-20121217"
+    },
+    "hr-time-20191121": {
+        "aliasOf": "hr-time-1-20191121"
+    },
```
tobie: Is there anyway this could be baked into the `switchWith` logic? It feels like this would solve a lot of our other related problems.

scripts/helper.js
ExE-Boss: Except&nbsp;that `hr‑time` is&nbsp;an&nbsp;alias for&nbsp;`hr‑time‑2`, not&nbsp;`hr‑time‑1`.

refs/w3c.json
ExE-Boss: Done in&nbsp;https://github.com/tobie/specref/pull/572/commits/3054155463a9e3d6d50e48db7ac484dec3b5dd6e.

scripts/helper.js
tobie: Oh! That's a good point. 

I'm still uneasy with what's going on, here.

overwrites/w3c.json
```diff
@@ -20,6 +20,13 @@
     "encoding": {
         "delete": true
     },
+    "hr-time": {
+        "renameTo": "hr-time-1",
+        "aliasOf": { "replaceWith": "hr-time-2" }
+    },
+    "hr-time-1": {
+        "href": { "replaceWith": "https://www.w3.org/TR/hr-time-1/" }
+    },
```
tobie: To clarify, here, we're trying to do:

1. Rename `hr-time` to `hr-time-1`
2. alias `hr-time` to `hr-time-2`
3. repace `hr-time-1` `href` with `https://www.w3.org/TR/hr-time-1/`

The way we're going about this seems a bit brittle, though, and not super clear.

I think this is essentially due to the fact that I chose an object to express what should essentially be an array of actions. Consider the following:

```json
[
  { "id": "hr-time", "action": "renameTo", "newId": "hr-time-1" },
  { "id": "hr-time", "action": "createAlias", "aliasOf": "hr-time-2" },
  { "id": "hr-time-1", "action": "replaceProps", "prop": "href", "value": "https://www.w3.org/TR/hr-time-1/" } 
]
```

This is uglier, I agree, but feels a lot less like it's relying on magic, no?

ExE-Boss: I&nbsp;feel&nbsp;like that&nbsp;should be&nbsp;its&nbsp;own&nbsp;PR.

tobie: Fair enough.

tobie:
Thanks for your contribution!

A new version of Specref updated with these changes should auto-deploy within minutes.

Thanks again!

PR #573, opened 2019-12-06
Refactor overwrites

tobie:
Fix a few bugs along the way


tobie:
OK, I'm merging this as the previous pull request has broken the autoupdate flow and this should address it.

Issue #574, opened 2019-12-09
Reporting API has moved; old URL still exists in SpecRef


clelland:
The Reporting API migrated out of incubation some time ago; however specref still returns the old URL for "REPORTING" (https://wicg.github.io/reporting/) rather than the adopted location (https://w3c.github.io/reporting/).

(This is mostly mitigated by GitHub's automatic redirection, but my bibilography entries are still wrong, and section-fragment-linking doesn't work through the redirect.)

PR #575, opened 2019-12-10
Move REPORTING spec from WICG to W3C.

clelland:
Fixes: #574

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -2351,6 +2351,20 @@
         "date": "2008",
         "isoNumber": "ISO/IEC 19757-2:2008"
     },
+    "REPORTING": {
+        "authors": [
+            "Douglas Creager",
+            "Ian Clelland",
+            "Mike West",
+            "Ilya Grigorik",
+            "Paul Meyer"
+        ],
+        "href": "https://w3c.github.io/reporting/",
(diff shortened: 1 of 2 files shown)


clelland:
I'm not 100% sure whether modifying wicg.json is okay here, but the spec is no longer returned from https://wicg.github.io/admin/biblio.json, so it should probably be removed.

marcoscaceres:
I'll remove it from the upstream version too... this file is pulled from the WICG admin repo.

marcoscaceres:
I misread the above... REPORTING is indeed gone from the upstream, so this should be correct.

marcoscaceres:
Having said that, this should show up automatically once published at as a FPWD on W3C's TR.

clelland:
Do you mean that REPORTING will move to w3c.json automatically? (And point to the ED?)

ExE-Boss:
`reporting` already&nbsp;exists in&nbsp;`w3c.json` as&nbsp;`reporting‑1`, so&nbsp;this&nbsp;needs to&nbsp;be&nbsp;converted into&nbsp;an&nbsp;alias, which&nbsp;is&nbsp;what I’m&nbsp;doing in&nbsp;<https://github.com/tobie/specref/pull/578>.

PR #583, opened 2020-02-04
chore(vscode): Add schema validation

ExE-Boss:
This&nbsp;makes it&nbsp;so&nbsp;that when&nbsp;viewing the&nbsp;files in&nbsp;[**Visual&nbsp;Studio&nbsp;Code**](https://code.visualstudio.com), schema&nbsp;validation is&nbsp;performed&nbsp;automatically.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,11 @@
+{
+    "json.schemas": [
+        {
+            "fileMatch": [
+                "/refs/*.json",
+                "/refs/**/*.json"
+            ],
+            "url": "/schemas/raw-reference.json"
+        }
+    ]
+}


tobie:
I think I'll need to be convinced why that belongs in the project.

ExE-Boss:
@connorshea articulated&nbsp;the&nbsp;reasoning a&nbsp;lot&nbsp;better in&nbsp;<https://github.com/mdn/browser-compat-data/pull/2905> than&nbsp;I&nbsp;could.

tobie:
Oh, I get how it works. It's just dedicated to a specific editor, and so that's the part that requires argumentation.  (I'm not saying I agree or disagree, I'm saying I need a convincing argument as to why that's OK.)

marcoscaceres:
As a VS Code user, I definitely wouldn’t say no. It’s a nice thing that doesn’t get in the way*.


* as long as it’s maintained.

tobie:
> *as long as it’s maintained.

That was my concern, too!

marcoscaceres:
I like that it just points to our existing schema - so there is not really much to maintain. @ExE-Boss are you willing to maintain this long term (5 years, say)  if we call on you.

ExE-Boss:
I’m willing to maintain this config file.

PR #595, opened 2020-03-26
Add reference to RFC6265bis for SameSite cookies

AutomatedTester:
This adds the reference to the draft specification that contains all of
information around same site feature of cookies.


tobie:
I think you’ll need to fix the JSON schema [here](https://github.com/tobie/specref/blob/b372a24b68b3d95949ff9b3a52e65c6a600c4061/schemas/raw-reference.json#L209). Is this a standard IETF notation, though?

marcoscaceres:
@AutomatedTester, just wondering how you want to proceed here?

AutomatedTester:
Looks like @foolip fixed it in the spec so we can close it out.

PR #600, opened 2020-05-22
Update info for Trusted Types spec

tidoust:
The spec moved to the Web Application Security Working Group and ED URL is no longer correct.

The need for the wicg.json overwrite is temporary while the entry gets removed from wicg/admin (I'll prepare a PR for that).

### DIFF
--- a/overwrites/wicg.json
+++ b/overwrites/wicg.json
@@ -1,0 +1,3 @@
+[
+    { "id": "TRUSTED-TYPES", "action": "delete" }
+]
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -2781,6 +2781,17 @@
         "publisher": "Apple, Inc.",
         "title": "TrueType™ Reference Manual"
     },
+    "TRUSTED-TYPES": {
+        "authors": [
+            "Krzysztof Kotowicz",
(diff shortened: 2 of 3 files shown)


marcoscaceres:
@tidoust we probably need to delete it from https://github.com/WICG/admin/blob/gh-pages/biblio.json#L122 also?

tobie:
I’m sorry that moving things from WICG to W3C is so painful. This is really something we should fix.

tidoust:
@marcoscaceres Yes, that's what I meant with "I'll prepare a PR for that". To avoid any discontinuity, update should take place in Specref before the entry gets deleted from WICG/admin/biblio.json.

... or we just don't care coz' discontinuity will be minimal and we just move ahead.

tidoust:
Thanks @tobie! A bit painful, indeed. Doesn't happen every day (and that's usually good news as in "making progress!" ;)). And PR to drop it from WICG biblio is there: https://github.com/WICG/admin/pull/110

tobie:
Yeah. I think the solution might be to run the import of both W3C and WICG in the same process and sharing the same output file. This would allow automating W3C priority imho.

PR #603, opened 2020-05-28
Add JSON-RPC to biblio.json

foolip:


refs/biblio.json
```diff
@@ -1805,6 +1805,15 @@
         "status": "ED",
         "publisher": "W3C"
     },
+    "JSON-RPC": {
+        "authors": [
+            "JSON-RPC Working Group"
+        ],
+        "href": "https://www.jsonrpc.org/specification",
+        "title": "JSON-RPC 2.0 Specification",
+        "date": "4 January 2013",
```
tobie, line 1814: This works too:

```suggestion
        "rawDate": "2013-01-04",
```

foolip: Let's go with that!

PR #610, opened 2020-06-13
Fix and update source URL in csswg.json

tidoust:
[Last update to the script](https://github.com/tobie/specref/commit/2d91325602246955ba2f68675989426c37e2e4da) moved the `biblio.ref` URL to HTTPS but did not update the source URL of all entries in `csswg.json`, which stuck to the HTTP one. This meant that the script could no longer integrate updates from `biblio.ref` (there haven't been any since then in practice, so nothing got missed).

This update switches the source URL to HTTPS, and updates it since it now is a redirect to https://drafts.csswg.org/biblio.ref

Note the `biblio.ref` file does not seem to be maintained anymore in practice so it's not entirely clear whether this should still considered to be a source, or whether the info in Specref actually becomes the source... see discussion at https://github.com/w3c/csswg-drafts/pull/5194#issuecomment-643397000

### DIFF
--- a/refs/csswg.json
+++ b/refs/csswg.json
@@ -1,12 +1,12 @@
 {
     "AAT-FEATURES": {
-        "source": "http://dev.w3.org/csswg/biblio.ref",
+        "source": "https://drafts.csswg.org/biblio.ref",
         "title": "Apple Advanced Typography Font Feature Registry",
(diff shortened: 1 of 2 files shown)


tobie:
Removing the script that pulls the data seems a reasonable thing to do.

PR #611, opened 2020-06-15
Drop W3C overwrites for hr-time

tidoust:
Info returned by W3C systems changed for hr-time, and the `renameTo` overwrite was creating an infinite loop (`hr-time-1` defined an alias of `hr-time-1`). The auto-update bot has been crashing in the past few days because of that.

### DIFF
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -8,10 +8,7 @@
   { "id": "notifications",     "action": "delete" },
   { "id": "dom",               "action": "delete" },
   { "id": "encoding",          "action": "delete" },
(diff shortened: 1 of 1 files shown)


tidoust:
@tobie With the fix, the update script works on my machine, but there should be a couple of updates that don't seem to appear so I suspect the bot is still stuck. Anything that may need to be restarted?

In particular, the but did not make the update script crash, but rather run forever. Can the bot be "still running"?

tobie:
> In particular, the but did not make the update script crash, but rather run forever. Can the bot be "still running"?

nope. :D 

The issue seems unrelated. Pushed a fix. We'll see how it goes.

PR #617, opened 2020-07-20
Rename Feature Policy spec to Permissions Policy.

clelland:
The spec name and URL have both changed, so this adds the new location, and
points the previous definitions of "feature-policy" and "feature-policy-1"
at it.

### DIFF
--- a/refs/w3c.json
+++ b/refs/w3c.json
@@ -34450,39 +34450,10 @@
         "title": "Testbed"
     },
     "feature-policy": {
-        "aliasOf": "feature-policy-1"
+        "aliasOf": "permissions-policy-1"
     },
     "feature-policy-1": {
-        "authors": [
-            "Ian Clelland"
-        ],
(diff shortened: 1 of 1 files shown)


clelland:
I'm not entirely certain about the validity of moving the version history into the new spec definition, so I'd appreciate a review before merging this.

@foolip @annevk FYI

tobie:
There’s currently an error blocking auto-update. This might very well be taken care of automatically as soon as auto-update is back up.

clelland:
Ahh, I figured it *should* be handled by the RSS feed reader, but I didn't see Permissions Policy anywhere in the file at all, and it looked like the original commit that added Feature Policy to w3c.json was done manually as well. So I guessed that this was the way :)

tobie:
You might be right, though.

marcoscaceres:
Merging, as this missing ref is causing breakage in specs.

PR #623, opened 2020-09-17
add ref for mathml-core

bkardell:


tobie (approved):
This LGTM. Not sure what the correct abbreviation is for CG drafts in the new process doc, though.

bkardell:
Is this in reference to the short name or a specific property in here?

tobie:
True status.

bkardell:
The process links to several CGs, among those reporting requirements/respec suggest CG-Draft might be appropriate

tobie:
Yeah, I guess we can always fix it later if need be.

PR #626, opened 2020-09-30
Update title of MathML Core to match actual spec title

tidoust:


tobie:
This might get overwritten by the bot if the data is wrong on W3C's side.

If so, please reopen a PR and fix in the overwrite files?

Thanks!

--tobie

PR #633, opened 2020-12-18
Update css-conditional-3

fantasai:
Handles spec rename + recent republication.


ExE-Boss (approved):

tobie:
That was fixed by the auto-update now that it was unblocked.

PR #641, opened 2021-03-15
Use datatracker.ietf.org URLs for IETF documents

dontcallmedom:
tools.ietf.org is no longer to be relied on per https://mailarchive.ietf.org/arch/msg/tools-discuss/oYrAxb3KayPzZ4SNB1DVZTDPPNo/

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -224,7 +224,7 @@
             "A. Phillips",
             "M. Davis"
         ],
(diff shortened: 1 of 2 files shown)


tobie:
That looks like it's only catching a very small number of IETF docs, no?

tobie:
OK, tested it at home after running an update, and this works beautifully.

Thanks for the pull request.

PR #651, opened 2021-04-22
Alias hr-time and resource-timing to latest draft

annevk:
I'm not entirely sure why this only needs to happen for these, but I guess that's the way it is.

cc @domenic @tabatkins

### DIFF
--- a/overwrites/w3c.json
+++ b/overwrites/w3c.json
@@ -8,7 +8,8 @@
   { "id": "notifications",     "action": "delete" },
   { "id": "dom",               "action": "delete" },
   { "id": "encoding",          "action": "delete" },
-  { "id": "hr-time",           "action": "createAlias", "aliasOf": "hr-time-2"},
+  { "id": "hr-time",           "action": "createAlias", "aliasOf": "hr-time-3"},
+  { "id": "resource-timing",   "action": "createAlias", "aliasOf": "resource-timing-2"},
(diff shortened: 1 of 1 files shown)


tabatkins:
I mean, probably more than this is needed, but these are at least known issues to be fixed.

tobie:
I don’t know either. Something wrong with the aliasing on the W3C side, I assume?

/cc @deniak

deniak:
We leave the decision to the WG to decide where the shortlink without the version number should point to.
For `hr-time`, it currently points to `hr-time-3` as shown in the API: https://api.w3.org/specification-series/hr-time?apikey=@apikey@

However, for `resource-timing`, there was no request to update the alias when `resource-timing-2` was published that's why it still points to `resource-timing-1`. @caribouW3, @plehegar, should we update the alias?

caribouW3:
There is a recent thread on what to do with resource timing
https://lists.w3.org/Archives/Public/public-web-perf/2021Apr/0008.html
My assumption is that level 1 will be simply dropped and 2 will become an unversioned living standard, but we can surely wait for the WG's decision.

annevk:
I guess that means I can remove the hr-time alias created here? I rather not remove the resource-timing alias as due to a culmination of things this ends up with duplicate references in Fetch.

deniak:
> I guess that means I can remove the hr-time alias created here? I rather not remove the resource-timing alias as due to a culmination of things this ends up with duplicate references in Fetch.

I think `hr-time` can indeed be removed. Hopefully the Web Perf WG will prefer to point to Level 2 soon and we can also remove `resource-timing`.

PR #655, opened 2021-04-27
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/specref/commit/011ce84b8fa4b7a0847fc9d9b5c2183e3c9032d7#commitcomment-50050420

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/specref/commit/57d13d6ddabda58a0321ca388186ac8972ef070b

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.


refs/etsi.json
```diff
@@ -11762,7 +11762,7 @@
     },
     "etsi-en-302-120": {
         "href": "http://www.etsi.org/deliver/etsi_en/302100_302199/302120/01.01.01_60/en_302120v010101p.pdf",
-        "title": "ETSI EN 302 120 V1.1.1 (2003-01): Services and Protocols for Advanced Networks (SPAN); Operation of the Bearer Independant Call Control (BICC) protocol with SIG.4/PNNI1.0/AINI signalling specification [ATM Forum specification AF-CS-VMOA-0146.000 (2000) modified]",
+        "title": "ETSI EN 302 120 V1.1.1 (2003-01): Services and Protocols for Advanced Networks (SPAN); Operation of the Bearer Independant [sic] Call Control (BICC) protocol with SIG.4/PNNI1.0/AINI signalling specification [ATM Forum specification AF-CS-VMOA-0146.000 (2000) modified]",
```
jsoref, line 11765: I've generally chosen to tag instances where the underlying content has a typo with `[sic]`.

refs/legacy.json
```diff
@@ -16,7 +16,7 @@
...
     "ICC42": "<a href=\"http://color.org/ICC1v42_2006-05.pdf\"><cite>Image technology colour management &#8212; Architecture, profile format, and data structure.</cite></a> International Color Consortium. Specification ICC.1:2004-10 (Profile version 4.2.0.0) With errata incorporated, 5/22/2006. URL: <a href=\"http://color.org/ICC1v42_2006-05.pdf\">http://color.org/ICC1v42_2006-05.pdf</a> ",
-    "ID-MGM": "Rundle, M. <a href = 'http://identityproject.lse.ac.uk/mary.pdf'><cite>International Data Protection and Digitial Identity Management Tools</cite></a> Internet Governance Forum 2006 - Athens, Privacy Workshop I. 31 October 2006. URL: http://identityproject.lse.ac.uk/mary.pdf ",
+    "ID-MGM": "Rundle, M. <a href = 'https://web.archive.org/web/20191227054302/identityproject.lse.ac.uk/mary.pdf'><cite>International Data Protection and Digital Identity Management Tools</cite></a> Internet Governance Forum 2006 - Athens, Privacy Workshop I. 31 October 2006. URL: https://web.archive.org/web/20191227054302/identityproject.lse.ac.uk/mary.pdf ",
```
jsoref, line 19: This repository occasionally uses IA links, and since I had to use IA to validate / correct this content, it seemed like a good idea to pin the URL as well (since the document is no longer available from the original source).

refs/ietf.json
```diff
@@ -45931,7 +45931,7 @@
     "rfc3074": {
         "rfcNumber": "RFC3074",
         "href": "https://datatracker.ietf.org/doc/html/rfc3074",
-        "title": "DHC Load Balancing Algorithm",
+        "title": "DHC [sic] Load Balancing Algorithm",
```
jsoref, line 45934: I'm curious about the story here. They're consistent in that the ipv6 version carries forward this spelling

refs/wg21.json
```diff
@@ -78374,7 +78374,7 @@
         "href": "https://wg21.link/lwg2304",
         "status": "C++14",
         "authors": [
-            "Joaquín M López Muñoz"
+            "Joaquín Mª López Muñoz"
```
jsoref, line 78377: Various names suffered from varying degrees of mojibake. When a name appeared both with fancy characters and either without or baked, I corrected to the fancy flavor.

My tooling likes to use lowercase ascii letters for commit messages, but those are really just for me to maintain an ordered sequence of tasks, I don't expect the commit messages to be retained when merged.

refs/wg21.json
```diff
@@ -103281,7 +103281,7 @@
         "title": "Toward Improved Optimization Opportunities in C++0x",
         "href": "https://wg21.link/n1664",
         "authors": [
-            "W. Brown,M. Paterno"
+            "W. Brown, M. Paterno"
```
jsoref, line 103284: There were a couple of places where I noticed missing spaces after `,`s. I've chosen to put them in, because ... they belong.

refs/ietf.json
```diff
@@ -93929,7 +93929,7 @@
         "href": "https://datatracker.ietf.org/doc/html/rfc6173",
         "title": "Definitions of Managed Objects for the Internet Fibre Channel Protocol (iFCP)",
         "authors": [
-            "P. Venkatesen, Ed."
+            "P. Venkatesan, Ed."
```
jsoref, line 93932: Some people's names are misspelled. Just because they're the editor of a document doesn't mean their name was spelled correctly.

In this case, I'm going with LinkedIn:

https://in.linkedin.com/in/prasanna-venkatesan-k-b5888a173

refs/legacy.json
```diff
@@ -89,7 +89,7 @@
...
     "SUITEB": "<a href=\"http://www.nsa.gov/ia/programs/suiteb_cryptography/\"><cite>NSA Suite B Cryptography</cite></a>. URL: <a href=\"http://www.nsa.gov/ia/programs/suiteb_cryptography/\">http://www.nsa.gov/ia/programs/suiteb_cryptography/</a> ",
     "SUS3": "Cathy Hughes. <a href=\"http://www.unix.org/single_unix_specification/\"><cite>The Single UNIX Specification, Version 3</cite></a>. 2004 URL: <a href=\"http://www.unix.org/single_unix_specification/\">http://www.unix.org/single_unix_specification/</a>",
-    "THEGRID": "Allen Hurlburt. <cite>Grid: A Modular System for the Design and Production of Newpapers, Magazines, and Books.</cite> Wiley. 1982. ISBN 0-471-28923-X.",
+    "THEGRID": "Allen Hurlburt. <cite>Grid: A Modular System for the Design and Production of Newspapers, Magazines, and Books.</cite> Wiley. 1982. ISBN 0-471-28923-X.",
```
jsoref, line 92: Some objects appear on the web in multiple flavors. This book is available with both spellings from Amazon, but the correct spelling can be seen on various places, including, https://www.si.edu/object/siris_sil_105907

refs/wg21.json
```diff
@@ -60878,7 +60878,7 @@
         "source": "https://wg21.link/index.json"
     },
     "EDIT814": {
-        "title": "[ﬁlesystems] add missing parens",
+        "title": "[filesystems] add missing parens",
```
jsoref, line 60881: The intention of this item was not to be ligatured, and there's no reason to make it harder for people to search for it.

jsoref (commented):
It's certainly possible that an intermediate or even primary source has errors (especially in names), but I claim that it would be better to generally use correct information, and if possible to get the intermediate/primary sources corrected, than it would be to continue to propagate incorrect information.

refs/wg21.json
```diff
@@ -114772,7 +114772,7 @@
         "title": "Minimizing Dependencies within Generic Classes for Faster and Smaller Programs",
         "href": "https://wg21.link/n2911",
         "authors": [
-            "D. Tsafrir, R. Wisniewksi, D. Bacon, B. Stroustrup"
+            "D. Tsafrir, R. Wisniewski, D. Bacon, B. Stroustrup"
```
jsoref, line 114775: > Dan Tsafrir, Robert W. Wisniewski, David F. Bacon, and Bjarne Stroustrup: Minimizing Dependencies within Generic Classes for Faster and Smaller Programs. ACM OOPSLA'09. October 2009.

https://www.stroustrup.com/papers.html

I'd also claim that it would be best to expand all authors instead of the inconsistent mix of abbreviated/unabbreviated, but I haven't actively done this.

refs/etsi.json
tobie: I'm not sure about this, tbh?

Maybe just leave those as is?

refs/wg21.json
```diff
@@ -13172,7 +13172,7 @@
         "source": "https://wg21.link/index.json"
     },
     "CWG2078": {
-        "title": "Name lookup of mem-initilizer-id",
```
tobie, line 13175: Are we sure about this?

refs/wg21.json
```diff
@@ -20370,7 +20370,7 @@
         "source": "https://wg21.link/index.json"
     },
     "CWG435": {
-        "title": "Change \"declararation or definition\" to \"declaration\"",
+        "title": "Change \"declaration or definition\" to \"declaration\"",
```
tobie, line 20373: That seems possibly a correction of a correction

refs/wg21.json
tobie: This should really be split up in two author fields.

refs/wg21.json
tobie: This should be split up, too.

tobie (commented):
There's a lot here.

Some of it is straightforward.

Some of it isn't as much.

And some of it might get overwritten at the next auto-update.

Not sure how to best move forward, tbh.

refs/wg21.json
jsoref: I mostly tried to skip those. But at some point I was letting Google Sheets make suggestions and guessing they were ok. Will drop.

refs/etsi.json
jsoref: Yeah, I'm happy to drop them. I did it mostly so I could have a temporary record of which things I had checked and confirmed were bugs upstream as opposed to incorrectly imported.

refs/wg21.json
jsoref: > 2078. Name lookup of mem-initilizer-id
> Section: 6.5.2  [class.member.lookup]     Status: NAD     Submitter: Richard Smith     Date: 2015-01-31
> The name lookup in a base-specifier and a mem-initializer differ in that the former ignores non-type names but the latter does not. When the mem-initializer-id is qualified, this can lead to surprising results:

The name indeed has the typo, the content does not.

No, I didn't check this one. (I did this late overnight.)

jsoref:
Obviously the easiest fixes are the stuff that's in your code (as opposed to the dataset).

I have no idea how your pipeline works, and I'd like to treat it as a black box. If someone can push fixes upstream so that they don't come back, that's best.

I'm still trying to release the version of my tool that I'm using (it's close, but I have two more things I want to fix, one hit in production overnight, and one which is just polish).

tobie:
Hi there, any chances of breaking this down into a few different pull requests where we tackle what's not external references first?

jsoref:
Sure. (Sorry, I've been trying to release 0.0.18, which I half did, and then ran into bugs in dependabot which is so much fun.)

jsoref:
For reference, the above is just rebasing spelling onto spelling-code -- so, temporarily there may be more commits, but once spelling-code merges, there will be fewer.

refs/wg21.json
jsoref: Are these the canonical forms? Or is this the output of a scraper?

jsoref:
This perl split the authors:

```perl
#!/usr/bin/env perl -pi
if ($a == 1) {
  if (/^\s*\]/) {
    $a = 0; next;
  }
  s/, Ed\./__ED/;
  if (/^(\s+").*,.*"/) {
    $lead = $1;
    s/,\s*(.)/",\n$lead$1/g;
  }
  s/__ED/, Ed./;
} elsif (!$a && /"authors": \[/) {
  $a = 1;
}
```

n.b. GitHub hates this commit.

jsoref:
@tobie / @marcoscaceres: would you want the split authors change standalone? (It comes with Eric Shepherd, because otherwise Shepherd would be split from Eric...)

or the internet archive: identityproject.lse.ac.uk/mary.pdf commit...?

Those are the only two that are easily splittable from this set. I mean, everything else could be split, but I don't think there's any particularly useful split beyond by file, and I doubt that helps much.

marcoscaceres:
I'm ok with the size of this change. The changes are pretty straight forward. 

@tobie?

tobie:
> I'm ok with the size of this change. The changes are pretty straight forward.

Most of these changes will get overwritten in the next 60 minutes by the auto update. 

We have to be more intentional if this is to be useful, which is why I suggested starting with just spelling mistakes outside of the data itself.

marcoscaceres:
> Most of these changes will get overwritten in the next 60 minutes by the auto update.

hehe, that was next question.... "wait! isn't all this automated?" 

> We have to be more intentional if this is to be useful, which is why I suggested starting with just spelling mistakes outside of the data itself.

Agree.

jsoref:
Yeah, I was worried that most of the content was not primary source, which is why I wasn't particularly eager to do the author split.

marcoscaceres:
Ok, so I guess we can salvage the changes to biblio.json and to legacy.json.

tobie:
@jsoref so I'm still interested to merge changes to `refs/biblio.json`, `refs/legacy.json`, the readme. the docs, and other non-automated parts of the data set.

Is this something you would be willing to look into?

jsoref:
Sure. What do you want me to do?

marcoscaceres:
thanks for extracting the PRs, @jsoref!

jsoref:
It's trivial for me to do so, I just need direction :-)

Is there a strategy for the other things? Are there upstreams we can poke? Are there tools we can improve?

marcoscaceres:
> Is there a strategy for the other things? 

I guess if you can see if those documents are on GitHub, then filing issues on the misspelled specs will get them fixed there. 

> Are there upstreams we can poke? 

As above... some are it's going to be difficult :( like the IETF ones. 

> Are there tools we can improve?

Yes... this could run exclusively when a PR is made to refs/biblio.js, for instance. At least, we could check those.

jsoref:
> Yes... this could run exclusively when a PR is made to refs/biblio.js, for instance. At least, we could check those.

I'd be quite happy to offer the action. It's trivial to exclude files (e.g. the three files remaining in this PR) using `excludes.txt`, or to only check files using `only.txt`.

marcoscaceres:
I think (thought please verify!) you can exclude also via the action syntax itself:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#excluding-paths

jsoref:
That doesn't do what you imagine. It controls _when_ an action runs, not what it does. And for more fun, it doesn't provide that content to the workflow, so you can't actually use it to do work, you have to re-engineer it :-o

marcoscaceres:
Fun! thanks @jsoref. I'll take a look at the corresponding PR soon.

PR #656, opened 2021-04-27
Fix test after alias changes for hr-time and resource-timing

deniak:


tobie (changes requested):
Thanks for doings this.

What we need to do instead here is manually alias those so we don’t remove references that folks could be relying on.

deniak:
> What we need to do instead here is manually alias those so we don’t remove references that folks could be relying on.

The problem is that these references break the duplicate tests as far I understand how this works.

deniak:
oh, I think I understand what you mean now. Let me give it a try

deniak:
@tobie, I jsut added the aliases

tobie (approved):

PR #658, opened 2021-05-04
Spelling sic

jsoref:
split from #655 

It's possible this will just be left as a PR for reference as opposed to being merged.


marcoscaceres (commented):
I don't agree with this change, tbh. If the title of the document is wrong, then that's the title. I don't think we should "[sic]" them.

tobie:
Closing in favor of a new PR focused on the non-automated parts of the program.

PR #675, opened 2021-06-24
Use default datatracker urls for HTTP RFCs

### LINKED ISSUE #672: Revert adoption of httpwg.org URLs
https://github.com/tobie/specref/pull/539 added a hardcoded map from a number of RFCs  to their (much more nicely rendered) equivalent in `httpwg.org` space, but without much discussion or background.

While the improved rendering is certainly valuable, I'm not sure if using a WG-specific domain without clear institutional backing is necessarily the right thing to do by default for links that may end up encoded in formally approved standards.

It's also impacting our [browser-specs](https://github.com/w3c/browser-specs/) tool which consumes data from specref to collect metadata about specifications, as we're [considering adding some of these RFCs](https://github.com/w3c/browser-specs/issues/280) to our list; we can easily override this on our end, but since the validity of the specref data was at least not obvious to us, we thought we would bring this up for discussion here first.

cc @tidoust @mnot

dontcallmedom:
close #672
revert #539

### DIFF
--- a/scripts/ietf.js
+++ b/scripts/ietf.js
@@ -40,48 +40,8 @@ request({
     helper.tryOverwrite(FILE);
 });
 
-// From https://httpwg.org/specs/
-// JSON.stringify([...document.body.textContent.matchAll(/(RFC\d{4}):/g)]
-//     .map(m => m[1].toLowerCase()).sort(), null, 4);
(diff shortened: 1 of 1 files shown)


dontcallmedom:
no longer pursuing this road

Issue #677, opened 2021-07-10
Remove WebIDL-2 alias (and associated dated aliases)


tabatkins:
SpecRef currently recognizes `WebIDL-1` and `WebIDL-2`, along with an unversioned `WebIDL`. The unversioned name points to the current spec in heycam's GH; the `-1` is an alias for the most recent /TR publication (from 2016!), and the `-2` is an alias for the unversioned name.

I have *no idea* where the `-2` version is coming from. I can't detect it in any of the files in this repo, and it doesn't appear to be in `tr.rdf` either.

This is causing some issues in my attempted switchover to [WebRef](https://github.com/w3c/webref/) as the definition source for Bikeshed. WebRef relies on [browser-specs](https://github.com/w3c/browser-specs), which only recognizes `WebIDL-1`, tagging the last /TR publication as the "release" URL and the heycam version as the "nightly" url. Definitions coming from WebIDL are thus claiming to be from the spec with shortname "webidl-1", which per SpecRef is the ancient /TR version, which is causing a lot of churn in the biblio sections of my test suite specs.

I can of course fix this on my end with manual intervention (I'm already just removing the ancient /TR url entirely from my anchor data source, as it's worthless and never worth linking to), but it would be good in general to have these SpecRef and browser-specs agree on this matter, and I suspect SpecRef is currently doing the weirder thing and should be the one to change.

I recommend that `WebIDL-1` be changed to an alias of `WebIDL`, so it'll default to the best information, and all the `WebIDL-2*` entries simply be removed.

marcoscaceres:
Agree... and maybe we can remove "WebIDL-1" entirely.

tobie:
Happy to merge a pull request. Our policy so far has been to never remove entries, just alias them instead, so if we want to remove entries here, we should start a separate conversation to discuss a policy change (and a process for deprecation/removal).

annevk:
So is the problem that it creates the alias before the WHATWG data is imported? Perhaps the overrides should happen after the imports?

How did we handle this for other specifications that moved?

PR #686, opened 2021-10-05
Account for WebIdl moving to WHATWG

tobie:
/cc @annevk


tobie:
Ugh. There's more stuff to cleanup in w3c.json.

tobie:
See also the conversation at: https://github.com/tobie/specref/issues/677

/cc @marcoscaceres

annevk:
What would the best course of action be here, also delete the -1 and -2 aliases? Does that need a policy discussion/decision now the document has moved? And if yes, is that documented somewhere since I'm not sure what it entails?

tobie:
@annevk: I'm going to confess having very little time to focus on this atm and don't want to be blocking y'all from moving forward with this. I'm happy with whatever you, @tabatkins, and @marcoscaceres agree to.

annevk:
Thanks and understood; I hope @marcoscaceres can chime in tomorrow with the best course of action here.

marcoscaceres:
I think we should do what @tabatkins suggested:

> I recommend that `WebIDL-1` be changed to an alias of WebIDL

And same with `WebIDL-2`. 

Let me see what's left over in w3c.json...

marcoscaceres:
Updating this PR locally... will send something up soon.

marcoscaceres:
ok, did the re-alias so no references are lost. 

Now all point to the right place ("WEBIDL"). 

@tobie, can you give it a once over to make sure it looks right?

marcoscaceres:
@annevk, can we add the version directly to "biblio.json" on the WHATWG side?

annevk:
On the WHATWG side we generate that file automatically and don't have a good way of adding additional data to it. We could probably put something together, but I'd rather we do it here somehow.

marcoscaceres:
Ok, I'll investigate.

marcoscaceres:
> Have you tried running an update over your code changes? It might well work?

No. Which scripts should I run?

tobie:
Not behind my desk atm, sorry for not being super responsive.

You want to just run the following lines: https://github.com/tobie/specref/blob/main/scripts/update-all#L39-L55

tobie:
Don’t run the whole script. Just copy/paste these lines.

marcoscaceres:
Argh... not quite right yet. Getting there tho.

PR #693, opened 2022-01-24
Add ISO 28500:2017

edsu:
This commit adds ISO 28500:2017 which is a specification used by the web
archiving community.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -5046,6 +5046,14 @@
         "publisher": "W3C Semantic Web Interest Group",
         "rawDate": "2006-02-01"
     },
+    "warc": {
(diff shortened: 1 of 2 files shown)


edsu:
@tobie did I do this right?

tobie:
That seems like another TC, though, no?

`/iso_jtc1_sc29.json` is pulling data automatically from a resource provided by @sandersaares.

Maybe just add this one to biblio.js for now?

edsu:
Thanks, I actually didn't fully understand the name of that file and added the WARC ISO there because I saw there were lots of other ISOs in it. How does it look now?

tobie:

Issue #706, opened 2022-04-20
Unable to reference new mediaqueries-3 REC


mattgarrish:
We had a reference in the epub 3.3 specification to `css3-mediaqueries` that lists the 2012 version as the current recommendation, but a new recommendation was published on April 5: https://www.w3.org/TR/mediaqueries-3/

I tried changing the reference to `mediaqueries-3` but this had no effect as it appear to be an alias.

I'm not familiar enough with how the specref database gets updated to say what the problem is, but I also noticed that the w3c.json file in the refs directory hasn't been updated in 16 days, which would be before the new recommended version was released.

Is there an issue with getting the latest data from W3C, or is the problem perhaps caused because an alias has already been assigned to the shortname used for the new recommendation? A little bit of each? Or am I way off base?

/cc @iherman

tobie:
> is the problem perhaps caused because an alias has already been assigned to the shortname used for the new recommendation?

Yes! That has broken the auto-update mechanism and I haven’t had the time to look into it.

mattgarrish:
Is it as easy as stripping this entry from the biblio.json file: https://github.com/tobie/specref/blob/6b142be84d2758249423aa81bbd6f8e4501b289e/refs/biblio.json#L2072

The danger I assume is that if someone really only wants to reference the 2012 version this would bump them forward, but is that a problem in practice?

frivoal:
> On Apr 22, 2022, at 2:17, Matt Garrish ***@***.***> wrote:
> Is it as easy as stripping this entry from the biblio.json file: https://github.com/tobie/specref/blob/6b142be84d2758249423aa81bbd6f8e4501b289e/refs/biblio.json#L2072 <https://github.com/tobie/specref/blob/6b142be84d2758249423aa81bbd6f8e4501b289e/refs/biblio.json#L2072>
> The danger I assume is that if someone really only wants to reference the 2012 version this would bump them forward, but is that a problem in practice?
> 

Shouldn’t be an issue, these two specs are normatively equivalent. The new one has editorial tweaks, and annotations indicating proposed (therefore not yet accepted as normative) corrections. Except for explicit comparison, (and in that case, they should be referred to with dated drafts), there’s no reason to specifically want to refer to the old one.

tobie:
Fixed in #709.

mattgarrish:
Thanks!

PR #707, opened 2022-04-26
Remove mediaqueries-3 alias from biblio file

mattgarrish:
Hopefully, this is all that's needed to fix the issue in #706


mattgarrish:
To get this to pass, I had to add aliases for all the mediaqueries-3 dated specs to the css3-mediaqueries dated spes. I followed what was done with the "mediaq" short name for this.

I also had to remove the reference to "mediaqueries-3" from ref-list.json. I assume this reference will be restored to its proper form once the W3C data is crawled again.

test/ref-list.json
```diff
@@ -44504,7 +44504,6 @@
     "MEDIAQUERIES-20170905",
     "MEDIAQUERIES-20200721",
     "MEDIAQUERIES-20211225",
-    "MEDIAQUERIES-3",
     "MEDIAQUERIES-3-20010404",
```
tobie, line 44507: So we can’t remove this test without breaking backwards compat. This needs to continue pointing to MQ3.

mattgarrish: If I put it back and not the alias, then I get the error:

>  1) Previous references are maintained
>       Reference [[MEDIAQUERIES-3]] is maintained.:
>     AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
>
>  assert.ok(current.get(key)[key])
>
>      at Context.<anonymous> (test\no-ref-loss.js:12:20)
>      at processImmediate (internal/timers.js:461:21)

Is that okay to ignore?

mattgarrish: Ah, okay. I'll leave it to the experts, then!

tobie: It isn’t expertise, it’s tinkering and patience.

mattgarrish: Sure, my tinkering has just hit a wall in that I don't understand what backwards compatibility is breaking, so I don't even know what else to tinker with. I only know enough to follow what a test run returns or I'd offer to help more.

tobie: Apologies for the lack of clarity. Specref's contract has always been that that although references might get updated, they're never removed. MEDIAQUERIES-3 should continue to point to something. The REC maybe?

mattgarrish: > MEDIAQUERIES-3 should continue to point to something. The REC maybe?

Won't this be restored once the new w3c data gets pulled into the w3c.json file? The new REC has the shortname mediaqueries-3, so I'm assuming it will result in an entry like:

```
"mediaqueries-3": {
   ...
   "versions": [
      "20200505": {
         ...
      }
   ]
}
``` 

which would only make for a temporary gap when the reference won't exist, but I'm not sure how else to work around that.

PR #719, opened 2022-09-02
Add window-controls-overlay spec ref to wicg.json

sonkkeli:


tobie:
Hi, I think you’d be better off getting it included upstream. Cc’ing @marcoscaceres to help with that.

tobie:
Here's the upstream I was referring to: https://github.com/WICG/admin/blob/gh-pages/biblio.json.

Closing on my end.

Issue #721, opened 2022-11-09
Where was manifest-incubations pulled from?


alancutter:
https://wicg.github.io/manifest-incubations/ was added in https://github.com/tobie/specref/commit/8954c99bc2047c1276d6ce9f90b58dfabe39494d by what looks like an automated process.
I'm interested in having https://wicg.github.io/web-app-launch/ added to specref, where does it need to live to be pulled in by this automated process?

alancutter:
https://github.com/tobie/specref#updating--adding-new-references discusses manually adding specs but fitting in with the automated process seems more ideal.

tobie:
If I recall correctly, this is the right file to add it to: https://github.com/WICG/admin/blob/gh-pages/biblio.json

PR #722, opened 2022-11-10
Add link to auto update script in README

alancutter:
This helps viewers determine where specs should be listed to get pulled into SpecRef automatically.

Fixes: https://github.com/tobie/specref/issues/721


tobie:
Thanks, Alan!

PR #731, opened 2023-02-05
Entries for CBD and SAX

gkellogg:
* CBD - Concise Bounded Description
* SAX: The Simple AP…I for XML.


gkellogg:
Removed the entry for "SAX", as this caused an odd `npm test` error. This might be related to https://github.com/w3c/respec/issues/4378#issuecomment-1416860913.

PR #738, opened 2023-03-07
Fix broken markup for MOZ-ICONS entry

tabatkins:
The `<a>` start tag wasn't closed. I slightly normalized the markup to match surrounding entries too. No text changes.


tobie:
Ideally, we just turn those into structured data when they need fixing (see [#147](https://github.com/tobie/specref/issues/147)).

tabatkins:
Right, I wouldn't mind doing so, but don't have the bandwidth right this moment. In the meantime, this has long actively caused issues for Bikeshed, because the parser I used would emit a complain about invalid markup to stdin just because something else had loaded the biblio file this reference was in.

tobie:
Oh, totally. I just wanted to make sure you were aware these were just legacy things.

tabatkins:
Given there's only a 100ish of them, that's a day's work when I have time. Do they all go into the `biblio.json` file, or do I need to decide which of the other files they might belong to? (I seem to recall the other files are more like overrides/additions to the "canon" biblio file?)

tobie:
Yeah, just putting them in `biblio.json` should be fine.

PR #741, opened 2023-04-21
Update FIPS refs.

davidlehn:
- Convert legacy FIPS refs to structured objects.
- Normalize FIPS refs formats.
- Add FIPS-186-4, FIPS-186-5.
- Change DSS to aliasOf FIPS-186-5.

I'm not sure what the best practice is for refs like "DSS".  It was the full FIPS-186-4 ref and this PR changes it to an alias.  It seems that since other docs may refer to specific FIPS publications, that "DSS" should be an alias to the latest FIPS ref and be updated every decade or so as new FIPS publications are released.

Also unsure what best practice is for what the href is.  These refs link to the PDF "Local Download" links NIST has, but they also have publication info pages with PDF links and more, and doi URLs for the PDF.  For example https://csrc.nist.gov/publications/detail/fips/186/5/final.


tobie:
> I'm not sure what the best practice is for refs like "DSS". It was the full FIPS-186-4 ref and this PR changes it to an alias. It seems that since other docs may refer to specific FIPS publications, that "DSS" should be an alias to the latest FIPS ref and be updated every decade or so as new FIPS publications are released.

That sounds reasonable indeed.
 
> Also unsure what best practice is for what the href is. These refs link to the PDF "Local Download" links NIST has, but they also have publication info pages with PDF links and more, and doi URLs for the PDF. For example https://csrc.nist.gov/publications/detail/fips/186/5/final.

It's hard to say what's best. As there's bi-directional linking anyway, I feel like what you picked is fine.

tobie:

PR #753, opened 2023-08-02
Resume browser-specs source, preserving entries

### LINKED ISSUE #752: Browserspec script is removing entries
The browser spec script is removing entries (currently `TC39-RESIZABLEARRAYBUFFER`) which (rightfully) breaks the build.

Script needs editing to keep those references around and potentially mark them as obsolete.

tidoust:
This update leverages the new `obsoletedBy` and `formerNames` properties reported by browser-specs to:
- track spec transitions over time,
- make sure that entries in `browser-specs.json` are preserved,
- and avoid the creation of ghost entries.

In particular, on top of creating `aliasOf` relationships when needed, the script now sets `isSuperseded` and `obsoletedBy` properties - or `isRetired` - for discontinued specs.

Next time the script runs, among other minor updates, it will flag `TC39-RESIZABLEARRAYBUFFER` as superseded and obsoleted by `ECMASCRIPT`, and flag `CLOSE-WATCHER` as superseded and obsoleted by `HTML`.

This should fix #752.

### DIFF
--- a/scripts/browser-specs.js
+++ b/scripts/browser-specs.js
@@ -30,9 +30,26 @@ function convertSpec(spec) {
     if (spec.nightly.repository) {
         ref.repository = spec.nightly.repository;
     }
+    if (spec.standing === 'discontinued') {
+        ref.isRetired = true;
+    }
+    if (spec.obsoletedBy) {
+        ref.obsoletedBy = spec.obsoletedBy;
+    }
     return ref;
 }
 
+
+function isInSpecref(id) {
+    if (updatedList[id]) {
+        return true;
+    }
+    var ref = refs[id];
+    if (ref && ref.source !== SOURCE) {
+        return true;
+    }
+    return false;
+}
 
 console.log("Updating", PUBLISHER, "references...");
 console.log("Fetching", SOURCE + "...");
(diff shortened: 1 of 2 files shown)


scripts/browser-specs.js
```diff
@@ -30,10 +30,27 @@ function convertSpec(spec) {
     if (spec.nightly.repository) {
         ref.repository = spec.nightly.repository;
     }
+    if (spec.standing === 'discontinued') {
```
tobie, line 33: Where does the standing prop come from?

tidoust: The [`standing`](https://github.com/w3c/browser-specs/#standing) property is maintained in browser-specs. That property is computed automatically based on the status of the underlying spec or based on the info returned by the W3C API in many cases. For most entries contributed by browser-specs to Specref, the `discontinued` value will rather be set manually because these specs rarely come with a machine-readable discontinued flag.

Rule in browser-specs is that, instead of deleting discontinued proposals, we will now preserve them and flag them as discontinued, optionally associating them with the list of specs that replaced them.

Similarly, we will now record changes of shortnames of a spec in browser-specs through a [`formerNames`](https://github.com/w3c/browser-specs/#formernames) property. This makes it possible to create/maintain `aliasOf` properties in Specref.

tobie: Get it. Thanks!

tobie: Sorry for the dumb question. I was trying to get back to you quickly and reviewed from my phone so was missing the broader context.

scripts/browser-specs.js
```diff
@@ -48,59 +65,110 @@ request({
...
-                } else {
-                    throw new Error("cannot override " + uppercaseId);
+            // No entry in Specref, let's add one
+            updatedList[uppercaseId] = convertSpec(spec);
+        });
+
+    // For continuity, add back aliases that we may have created earlier on,
+    // unless the targeted entry no longer exists (that should never happen
+    // though). Also add back entries that used to exist but that are now known
+    // under another name.
```
tobie, line 114: FWIW, it's perfectly fine to throw an error in such cases, ideally that's clear enough to explain why this is weird and should never happen. Throwing in the update script only means that auto-update is broken. Nothing else.

tobie (approved):

scripts/browser-specs.js
tidoust: Well, that seemed like a very good question to me! And I hope my reply did not come through as an RTFM one, I was just trying to provide more context ;)

PR #756, opened 2023-09-22
Move CLDR to https

aphillips:
The link checker complains that http is a permanent redirect and using https is just good hygiene anyway.


tobie:
sorry, totally missed this. Should be deployed within minutes.

PR #757, opened 2023-10-03
Change ETSI SmartM2M reference ontology URL from http to https URL. 

egekorkan:
Without this fix, there are warnings in W3C tools such as the [link checker](https://validator.w3.org/checklink).


tobie:
Thanks for the PR

PR #763, opened 2023-11-14
Update based on latest browser-specs

dontcallmedom:
not sure why the automated version doesn't run

### DIFF
--- a/refs/browser-specs.json
+++ b/refs/browser-specs.json
@@ -43,6 +43,17 @@
         "source": "https://raw.githubusercontent.com/w3c/browser-specs/web-specs%40latest/index.json",
         "repository": "https://github.com/w3c/audio-session"
     },
+    "AV1-AVIF": {
+        "href": "https://aomediacodec.github.io/av1-avif/",
+        "title": "AV1 Image File Format (AVIF)",
+        "status": "Editor's Draft",
+        "publisher": "Alliance for Open Media",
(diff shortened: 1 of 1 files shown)


refs/browser-specs.json
```diff
@@ -79,7 +90,7 @@
     "CLOSE-WATCHER": {
         "href": "https://wicg.github.io/close-watcher/",
         "title": "Close Watcher API",
-        "status": "Editor's Draft",
+        "status": "Living Standard",
```
dontcallmedom, line 93: hmm... browser-specs has it as discontinued and replaced by html, which this doesn't represent

tidoust: The spec is correctly represented as discontinued and replaced by html in Specref through other properties a few lines below this one:

```json
        "obsoletedBy": [
            "HTML"
        ],
        "isSuperseded": true
```

The status change is just an artefact of the way browser-specs maintains its info, even for discontinued specs (it follows the redirect, and bumps into the HTML spec whose status is "Living Standard"). The `"status"` property does not mean much for discontinued specs in any case.

tidoust (approved):
That looks good to me and I don't know why the automation did not run either. The script is called in `run-all` in particular, and running that script updates `browser-specs.json` as in this pull request...

dontcallmedom:
@tobie in case you get a chance to check why this isn't running in the automated context…

PR #765, opened 2023-11-23
Add URLPATTERN to whatwg.json and drop it from browser-specs.json

tidoust:
There seems to remain a transition hiccup when a WHATWG spec that was in `browser-specs.json` starts appearing in `whatwg.json`. Last update to `browser-specs.json` updated the `"URLPATTERN"` entry to flag it as an alias of `"WHATWG-URLPATTERN"`, which seems correct as there was no `"URLPATTERN"` entry in `whatwg.json`.

However, that update means that the `fetch-refs.js` script, applied to WHATWG specs, now wants to create a `"URLPATTERN"` entry in `whatwg.json`. That creates a duplicate, which prevents further scripts from running.

This update drops the alias entry from `browser-specs.json` and adds the right entry to `whatwg.json` instead. This should allow builds to resume.

### DIFF
--- a/refs/browser-specs.json
+++ b/refs/browser-specs.json
@@ -1733,9 +1733,6 @@
         "source": "https://raw.githubusercontent.com/w3c/browser-specs/web-specs%40latest/index.json",
         "repository": "https://github.com/WICG/ua-client-hints"
     },
-    "URLPATTERN": {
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):
thanks for digging into this!

PR #770, opened 2023-12-05
Make the MSE alias target media-source-1

tidoust:
Previous update missed the "MSE" alias that was hiding in `biblio.json`. The alias targets `media-source`, which means it now targets `media-source-2`. That's good but means that Specref can no longer resolve `MSE-YYYYMMDD` refs to the dated level 1 versions.

In turn, that makes tests fail. +1 for the consistency tests. -1 for the ability to alias names in different places!

This update creates the missing references as aliases explicitly, fixing tests.

I guess this shows that a more generic mechanism might be needed... but is not trivial ;)

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -2377,6 +2377,39 @@
     "MSE": {
         "aliasOf": "media-source"
     },
+    "MSE-20130129": {
+        "aliasOf": "media-source-1-20130129"
+    },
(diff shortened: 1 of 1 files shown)


tidoust:
Ah, but tests will only pass when the `w3c.json` file gets updated... I'll push the update manually to make sure that I'm not breaking something else again...

PR #771, opened 2023-12-14
Update VSS reference and introduce MQTT

tguild:
GENIVI relaunched as COVESA and continues to maintain VSS, presently at v4.0

OASIS' MQTT wasn't in specref so added that as well


tguild:
VSS entry threw an error, retrying

PR #780, opened 2024-02-27
Make ISOBMFF entry an alias of iso14496-12

tidoust:
Via https://github.com/w3c/encrypted-media/issues/523

The URL to the spec is no longer valid because the ISOBMFF spec is no longer publicly available. On top of that, it seems better to redirect to the actual entry in the ISO list of specs which is maintained automatically.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -2038,12 +2038,7 @@
         "href": "https://www.iso.org/standard/69084.html"
     },
     "ISOBMFF": {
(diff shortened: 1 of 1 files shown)


dontcallmedom (commented):
does it matter that the existing reference seemed to have been specifically to the 2015 version and that the latest is 2022?

dontcallmedom (approved):

PR #798, opened 2024-06-20
Fix WEBAC's title (as defined in reference) and href (to canonical)

csarven:
This leaves the `WEBAC` entry as is but corrects the `title` and `href` values.

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -3317,8 +3317,8 @@
         "aliasOf": "webdatabase"
     },
     "WEBAC": {
-        "title": "Web Access Control Ontology",
-        "href": "https://www.w3.org/ns/auth/acl",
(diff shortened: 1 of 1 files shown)


csarven:
Pinging @ajs6f as the original contributor for the WEBAC entry in case you'd like to raise any concerns. I believe this was referenced in Fedora API but the HTML for the [reference section](https://fcrepo.github.io/fcrepo-specification/#references) wasn't generated.

Hard to say whether anything else is referencing `WEBAC` but if it is only Fedora API, and there is the possibility to change, I suggest using something like `ACL` (probably best) or `ACL-ONTOLOGY` or `AUTH-ACL`.

The WAC specification references the ontology under [Namespaces](https://solidproject.org/TR/wac#namespaces) in text but does not link. It could use WEBAC or another key.

PR #802, opened 2024-07-29
Add two IANA registries

ralfhandl:
Add two more IANA registries, following the format of existing registries:
- HTTP Authentication Schemes
- HTTP Status Codes


ralfhandl:
@tobie please review and merge

PR #808, opened 2024-08-05
Update WICG refs with isolated contexts

shangl:


shangl:
@tobie , friendly ping :). PTAL.

tobie:
Why not upstream?

tobie:
i.e. here: https://github.com/WICG/admin/blob/gh-pages/biblio.json

GrapeGreen:
It's actually already there (see `ISOLATED-CONTEXTS`). We thought that adding this to the specref database could also make sense :)

The automatic update script for pulling refs from wicg into here seems to be broken.

tobie:
Fixed in https://github.com/tobie/specref/commit/6e0f94c13b8cd525bd74c5675ab8f2af4c910231

PR #809, opened 2024-08-08
JSON Reference draft

ralfhandl:
Adding [JSON Reference draft](https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03) which is referenced by
* [JSON Schema Draft 04](https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-04#section-10.2)
* OpenAPI [2.0](https://spec.openapis.org/oas/v2.0.html#reference-object), [3.0.0](https://spec.openapis.org/oas/v3.0.0.html#reference-object) - [3.0.3](https://spec.openapis.org/oas/v3.0.3.html#reference-object)

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -5150,6 +5150,17 @@
     "json-patch-20130401": {
         "aliasOf": "rfc6902"
     },
+    "json-reference": {
+        "authors": [
+            "Paul Bryan",
+            "Kris Zyp"
+        ],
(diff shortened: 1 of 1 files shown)


ralfhandl:
@handrews please check that the new entry is correct

handrews:
@ralfhandl sorry I did not see this earlier- yes, it's correct.

PR #819, opened 2024-10-11
Make "ttaf1-dfxp" an alias of "ttml1"

tidoust:
The "ttaf1-dfxp" shortname no longer exists and redirects to "ttml1" in W3C. This update makes it an alias of "ttml1" in Specref too.


dontcallmedom (approved):
I've confirmed this was the old shortname e.g. https://www.w3.org/TR/2010/REC-ttaf1-dfxp-20101118/

Issue #829, opened 2024-10-21
Need to document W3C entries obsoleted by WHATWG specs


tidoust:
The `tr.rdf` file (same for the W3C API) only knows about W3C specs. As such, it does not contain the information needed to fill the `obsoletedBy` properties of W3C specs that got superseded by WHATWG specs. When the shortname stayed the same, the data essentially moved to `whatwg.json`, which seems fine, but when the shortname changed in the process, the relationship gets lost in data. Main missing links:
- `cors` -> `fetch`
- `html5` entries -> `html`
- `xmlhttprequest` entries -> `xhr`

JohnRiv:
Another example that will need to be handled: `2dcontext` -> `html`

I'm not sure when https://www.w3.org/TR/2dcontext/ started redirecting to HTML, but I just noticed it does now (and it has its own section in it: https://html.spec.whatwg.org/multipage/canvas.html#2dcontext)

PR #830, opened 2024-10-21
Drop `access-control` from biblio.json, now handled in w3c.json

tidoust:
`access-control` was marked as an alias of `fetch`. That's not incorrect, but history-wise, `access-control` is more an alias of `cors`, which got superseded by `fetch`. That history wasn't properly recorded in the W3C database. It now is, so W3C update script is currently stuck because it wants to add an `access-control` entry that targets `cors`, duplicating the entry in `biblio.json`. This update drops the entry from `biblio.json`, so that updates may resume.

(The superseding relationship between `cors` and `fetch` is not properly captured yet but that's orthogonal to the problem at hand, see #829).

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -4028,9 +4028,6 @@
         "status": "Final",
         "title": ".ZIP File Format Specification"
     },
-    "access-control": {
-        "aliasOf": "FETCH"
-    },
     "bibo": {
         "authors": [
             "Bruce D'Arcus",


tidoust:
... And the update also adds the entry that the W3C update script would add to `w3c.json` to avoid a temporary hiccup where the entry would have disappeared.

deniak (approved):

PR #834, opened 2024-11-29
DCAT versions and aliases

dr-shorthair:
Renamed vocab-dcat --> vocab-dcat-1 

Added 
- DCAT --> vocab-dcat
- vocab-dcat --> vocab-dcat-3

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -933,6 +933,9 @@
         "publisher": "DCMI",
         "status": "DCMI Recommended Resource"
     },
+    "DCAT": {
+        "aliasOf": "vocab-dcat"
+    },
     "DCDSP": {
         "authors": [
             "Mikael Nilsson"


dr-shorthair:
Check is failing. 
Probably because w3c.json is updated automatically by W3C and should not be manually editted. 

However, there is inconsistency with the set of vocab-dcat* entries. 
The entry labelled "vocab-dcat' actually describes DCATv1 which also has an entry labelled "vocab-dcat-1", and the obsoleting labels are incomplete. 
I've tried to disentangle it, but (unsurprisingly) my manual edits to w3c.json are rejected. 

How can we get w3c.json fixed? 

@tidoust is this on you? 

Meanwhile, I'll make a simpler change only to biblio.json

tidoust:
Build fails because there's a strict "any entry that ever existed must continue to exist" rule in place, and the update would drop entries that used to exist. For example, `vocab-dcat-20120405` used to exist, but if you make `vocab-dcat` an alias of `vocab-dcat-3`, that entry will no longer exist because there is no `20120405` version under `vocab-dcat-3`.

There is no easy way to disentangle this right now. This is triggered by the fact that DCAT started with using a level-less shortname `vocab-dcat`, but then switched to levels. In practice, there are two `vocab-dcat`:
1. The `vocab-dcat` that now is `vocab-dcat-1`. That explains the clunky entries in Specref.
2. The `vocab-dcat` that represents the series of specs and currently targets level 3. Specref does not undestand series. Essentially, what you're trying to achieve cannot be done today in Specref. There's no good way to say "vocab-dcat used to be level 1, but now is level 3".

Once the W3C source has switched to the W3C API (PR #828), it will be possible to consider adding support for specification series. That update will also disentangle a couple of entries à la `vocab-dcat`. In the meantime, I would suggest to leave things as is.

Issue #811, opened 2024-08-20
Consider supporting entries for specification series


tidoust:
Via discussion in https://github.com/tobie/specref/pull/810#discussion_r1722985394

Various specs in W3C come in levels that belong to the same series. Both levels and series have shortnames in W3C, but the series shortname is more to be seen as a way to redirect to the "current" level (which evolves over time). For example: `css-fonts-3`, `css-fonts-4`, `css-fonts-5` all belong to the `css-fonts` series. As the time of writing, the current spec is `css-fonts-4`.

Levels may be created after a first iteration on a spec. For example, `media-source-1` was known as `media-source` until level 2 got created. `media-source` is now the shortname of the underlying series, the current spec being `media-source-2`.

Specref does not have the concept of series. That means no entry for `css-fonts`, and no easy way to "follow" the current specification in the latter case, requiring clunky workarounds, as done in #810. It would be great to support series shortnames more directly in Specref. 

The end goal would be to make it easy to reference a spec by:
1. its series shortname, such as `css-fonts`, in which case the reference goes to the latest version of the current level in the series.
2. its shortname, such as `css-fonts-5`, in which case the reference goes to the latest version of that level.
3. a dated version, such as `css-fonts-4-20240201` in which case the reference goes to the specified version. The dated version typically includes the level, we should not need to create or use a `css-fonts-20240201` entry!

Adding support for series requires switching the W3C update script to the W3C API, because [`tr.rdf`](https://www.w3.org/2002/01/tr-automation/tr.rdf) does not know anything about series either.

Note the need to preserve past dated versions in the case where the level gets introduced afterwards (e.g. `media-source-20140717` needs to exist even though the `media-source` spec is now known as `media-source-1`). Also note that some specs start without a level, then get a level, and then get back to a no-level mode...

(I don't think the notion of specification series exists in other SDOs, although that could perhaps be useful to track versions of ETSI and ISO specs longer term)

tobie:
My suggestion would be to have the dated versions tied to the unversioned entry, so: `FOO-20240201`, not `FOO-2-20240201`.

tidoust:
> My suggestion would be to have the dated versions tied to the unversioned entry, so: `FOO-20240201`, not `FOO-2-20240201`.

[...]

PR #835, opened 2024-11-29
DCAT alias

dr-shorthair:
Alias
DCAT --> vocab-dcat-3


dr-shorthair:
@tidoust this one should be safe

tidoust:
Tests do pass but note that's the same issue as in #834 that just gets postponed to a future date.

The `vocab-dcat-3` spec creates entries such as `vocab-dcat-3-20201217`. If `DCAT` is made an alias to `vocab-dcat-3`, then that means entries such as `DCAT-20201217` get created. If, later on, a level 4 gets worked on and you want to update `DCAT` to target `vocab-dcat-4` instead of `vocab-dcat-3`, that won't work because `DCAT-20201217` would then no longer exist. Again ,the right fix would be to have support for specification series (raised in #811) and the general guideline is "use references with levels" in the meantime.

I'm not trying to block this. Other entries got created using that model before, including some I fear I contributed. But that does not fit nicely with the way Specref models things for now.

dr-shorthair:
So are you saying that there is currently no way to refer to version-free DCAT using specref. 
And no way to fix that. 

I'm editing the new edition of the W3C SSN Ontology and we wish to draw attention to the parallel between ObservationCollection and dcat:Dataset. That DCAT class is in all editions of DCAT, so the reference should reflect that. 

[vocab-dcat] links to a version that is clearly marked 'obsolete' so I don't want to use that. 
I guess I'll have to add a local entry in the config.js instead. That seems unsatisfactory.

tidoust:
> So are you saying that there is currently no way to refer to version-free DCAT using specref.

That's correct. For example, there's no way to reference `css-color`. You have to choose between `css-color-3`, `css-color-4`, `css-color-5`, or `css-color-6`: https://www.specref.org/?q=css-color

> And no way to fix that.

No *good* way. There's always a way to force entries into Specref. If `DCAT` is made an alias of `vocab-dcat-3` then, when time comes to update the link to `vocab-dcat-4`, it is possible to force the creation of previous version entries in Specref, meaning to define entries such as:

```json
"DCAT-20201217": {
  "aliasOf": "vocab-dcat-3-20201217"
}
```

This will have to be done for all dated versions of `vocab-dcat-3`. That's the current workaround and there are entries done like this in the `w3c.json` file.

> I guess I'll have to add a local entry in the config.js instead. That seems unsatisfactory.

That seems like a good solution to solve your immediate problem. I suspect I'll look into adding support for specification series to Specref in a not too distant future because I need it as well. But that depends on other pending updates.

Issue #843, opened 2025-01-02
Open Font Format URL has changed


JohnRiv:
I noticed the current URL for `OPEN-FONT-FORMAT` (https://standards.iso.org/ittf/PubliclyAvailableStandards/c052136_ISO_IEC_14496-22_2009%28E%29.zip) returns an HTML page with this message:

> The page you're looking for has either moved or no longer exist.

I was able to find the updated URL in the ISO Standards website here: https://www.iso.org/standard/74461.html

Interestingly, that URL isn't in SpecRef yet, but the 2 updated versions of it (from amendments that were made) do exist in [refs/iso_jtc1_sc29.json](https://github.com/tobie/specref/blob/main/refs/iso_jtc1_sc29.json):
- **iso14496-22-2019-amd1-2020:** https://www.iso.org/standard/78865.html
- **iso14496-22-2019-amd2-2023:** https://www.iso.org/standard/82487.html

To address this, 2 things should probably happen:
1. Create a new entry for the new URL before the 2 amendments were added, i.e.:
  - **iso14496-22-2019:** https://www.iso.org/standard/74461.html
2. Alias `OPEN-FONT-FORMAT` to `iso14496-22-2019`

That would likely involve updates to `refs/csswg.json` and `refs/iso_jtc1_sc29.json` at a minimum.

JohnRiv:
I'm open to other ideas on how to handle this. I plan to do a PR tomorrow on how I think this should be addressed and welcome corrections.

JohnRiv:
I looked into this a bit more today and have an updated approach:

Assuming my PR to the csswg-draft repo is merged (since that is used as the source of the OPEN-FONT-FORMAT entry), I'll then create a PR to add `iso14496-22-2019` as its own separate entry in [refs/iso_jtc1_sc29.json](https://github.com/tobie/specref/blob/main/refs/iso_jtc1_sc29.json). (I considered creating it as an alias to `OPEN-FONT-FORMAT`, but that would eventually no longer be valid when that standard is updated, which is apparently [in the works](https://www.iso.org/standard/87621.html))

svgeesus:
@vlevantovsky is this change of url expected for OFF?

svgeesus:
I checked and [Incremental Font Transfer](https://w3c.github.io/IFT/Overview.html#biblio-iso14496-22) links to https://www.iso.org/standard/87621.html which is the current Draft Intentional Standard.

vlevantovsky:
[...]

PR #844, opened 2025-01-07
Creating iso14496-22-2019 & aliasing OPEN-FONT-FORMAT to that

JohnRiv:
The `href` update to the Open Font Format as well as the creation of the `iso14496-22-2019` entry for that specific version addresses #843 and is based on the conversation there.

I removed the `source` reference due to the comments in https://github.com/w3c/csswg-drafts/pull/11433 that the CSS WG's `biblio.ref` isn't used by anything any longer. I'll create a separate issue to address completely removing SpefRef's reference/usage of https://drafts.csswg.org/biblio.ref

### DIFF
--- a/refs/csswg.json
+++ b/refs/csswg.json
@@ -94,10 +94,7 @@
         "href": "https://www.w3.org/International/articles/typography/justification"
     },
     "OPEN-FONT-FORMAT": {
-        "source": "https://drafts.csswg.org/biblio.ref",
-        "title": "Information technology — Coding of audio-visual objects — Part 22: Open Font Format",
-        "publisher": "International Organization for Standardization.",
-        "href": "http://standards.iso.org/ittf/PubliclyAvailableStandards/c052136_ISO_IEC_14496-22_2009%28E%29.zip"
+        "aliasOf": "iso14496-22-2019"
     },
     "OPENTYPE": {
         "source": "https://drafts.csswg.org/biblio.ref",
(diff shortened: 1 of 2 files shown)


refs/csswg.json
```diff
@@ -94,10 +94,7 @@
         "href": "https://www.w3.org/International/articles/typography/justification"
     },
     "OPEN-FONT-FORMAT": {
-        "source": "https://drafts.csswg.org/biblio.ref",
-        "title": "Information technology — Coding of audio-visual objects — Part 22: Open Font Format",
-        "publisher": "International Organization for Standardization.",
-        "href": "http://standards.iso.org/ittf/PubliclyAvailableStandards/c052136_ISO_IEC_14496-22_2009%28E%29.zip"
+        "aliasOf": "iso14496-22-2019"
```
tidoust, line 97: Shouldn't `scripts/csswg.js` be disabled first? I suspect all will go fine but the script might be tempted to override this update next time it runs with the info it still finds in the outdated CSS WG source.

JohnRiv: I originally thought the same, but then I tested locally and it doesn't happen. Looking at the script:

https://github.com/tobie/specref/blob/5e7c635cc904fed15bbc746996c1ca9ef1cfa4a3/scripts/csswg.js#L66

it appears it's because I removed the `"source"` attribute for OPEN-FONT-FORMAT as part of this that it is not overwritten

tidoust (approved):

PR #850, opened 2025-01-27
Run a full W3C sync

tidoust:
This update is the result of running a full synchronization with the W3C API (instead of an incremental one, done locally because it takes ~2h30). Since this is the first time that a full sync is done, many entries get updated at once. The updates bring more consistencies between the entries (for example, authors appear everywhere, instead of just in versions), add missing versions for a number of specs, and updates the `source` field of all entries to target the W3C API.

This last update is going to be useful to add support for specification series, as discussed in #811 (the `source` field can then serve as identifier from a W3C API perspective).


tidoust:
@dontcallmedom I see diff is too large for GitHub. Reviewing updates individually would make little sense in any case. Tests ensure that no entry and no version were lost in the process!

dontcallmedom (approved):
trusting the tests then :)

PR #856, opened 2025-03-15
json-schema: reference latest IETF drafts

ralfhandl:
Update entries for json-schema-2020-12 and json-schema-validation-2020-12: reference latest IETF draft "-01" and update rawdate

### DIFF
--- a/refs/biblio.json
+++ b/refs/biblio.json
@@ -5304,9 +5304,9 @@
                     "Ben Hutton",
                     "Greg Dennis"
                 ],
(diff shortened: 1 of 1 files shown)


ralfhandl:
@relequestual This references the latest IETF drafts "-01" for JSON Schema Core and JSON Schema Validation, which are also the ones currently available at
* https://json-schema.org/draft/2020-12/json-schema-core
* https://json-schema.org/draft/2020-12/json-schema-validation

A valid alternative would be to use these json-schema.org URLs instead of the IETF ones.

Please check and provide feedback.

Thanks in advance!

tidoust (approved):

PR #868, opened 2025-07-03
Refs for W3C SSN Ontology 2023 Edition

dr-shorthair:


refs/biblio.json
```diff
@@ -3330,6 +3380,41 @@
     "SSML10": {
         "aliasOf": "speech-synthesis"
     },
+    "SSNX": {
```
tidoust, line 3383: This seems to be an article that describes the SSNX ontology. Shouldn't that rather target the [SSNX ontology itself](https://www.w3.org/2005/Incubator/ssn/ssnx/ssn)?

PR #891, opened 2025-10-03
Make `epub-a11y-111` an alias of `epub-a11y-12` (same with `-tech`)

### LINKED ISSUE #890: Version change in EPUB specs causing errors
We changed version numbers on the EPUB accessibility and techniques documents from 1.1.1 to 1.2 mid-revision yesterday (epub-a11y-111 -> epub-a11y-12 and epub-a11y-tech-111 -> epub-a11y-tech-12)  but it looks like this is causing a problem for auto updating: https://github.com/tobie/specref/actions/runs/18223440526/job/51888523487#step:4:850

As far as I know we did everything right on the W3C side, as the new versions are published and the histories now bridge the 1.1.1 and 1.2 working drafts. I don't know what the duplicate 1.2 URL errors are about as they are all for what should only be 1.1.1 drafts. Is it an aliasing problem, perhaps?

At any rate, if this is something we need to fix in the published drafts, please let me know.

/cc @iherman @deniak

tidoust:
Version was renamed on W3C side, which confuses the W3C script. Fixes #890.

The update also brings other changes to `w3c.json`. Not needed to unlock automatic updates, but that simplified patching the data!


tidoust:
To clarify, the only manual changes are lines [76932](https://github.com/tobie/specref/pull/891/files#diff-9f7fa13550d96ec3f1ba161c2481f1c9e646d9d2d9bdbcddfda607242f293d24R76932) and [79015](https://github.com/tobie/specref/pull/891/files#diff-9f7fa13550d96ec3f1ba161c2481f1c9e646d9d2d9bdbcddfda607242f293d24R79015). Other updates are just the result of running the `w3c.js` script.

PR #931, opened 2026-05-31
Adjust IETF logic for new RFC Editor site

tidoust:
IETF released a new RFC Editor site last week. The `rfc-index.xml` file is no longer available (related PR: https://github.com/ietf-tools/red/pull/268) at the URL that the IETF script was using. A similar index file is available from:
  https://www.ietf.org/rfc/rfc-index.xml

The new RFC Editor site also changes all URLs
from `http://www.rfc-editor.org/rfc/rfcXXXX`
to `http://www.rfc-editor.org/info/rfcXXXX/`

This update adjusts the IETF script to the new site and index.

The new index file no longer pads rfc numbers by default, e.g., it has `RFC1` instead of `RFC0001`. The script created an entry for `rfc0001` and an alias for `rfc1`. To better align with the new RFC index, it now does the opposite: it creates an entry for `rfc1` and an alias for `rfc0001`.

The `obsoletes`, `updatedBy, `obsoletedBy` properties now also target unpadded RFC numbers.

(The script could rather be tweaked to do the opposite and keep padded RFC numbers as the main reference, but the code would be more convoluted and both the unpadded and padded entries continue to exist for consumers in any case.)

One of the IETF overwrites needed to be adjusted as well. The other one was no longer needed and was dropped.

Note that the raw schema had to be relaxed to allow for an unpadded `rfcNumber`. This should not create any issue as that property is not exposed by the API.

### DIFF
--- a/overwrites/ietf.json
+++ b/overwrites/ietf.json
@@ -1,4 +1,3 @@
 [
-  { "id": "rfc1384", "action": "replaceProp", "prop": "obsoletedBy", "value": ["rfc1617"] },
-  { "id": "rfc0200", "action": "replaceProp", "prop": "obsoletedBy", "value": ["RFC-INDEX"] }
+  { "id": "rfc200", "action": "replaceProp", "prop": "obsoletedBy", "value": ["RFC-INDEX"] }
 ]
--- a/schemas/raw-reference.json
+++ b/schemas/raw-reference.json
@@ -205,7 +205,7 @@
                 "rfcNumber": {
                     "description": "Number of IETF RFC",
                     "type": "string",
-                    "pattern": "^(BCP|RFC)\\d{4}$"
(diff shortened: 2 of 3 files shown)


deniak (approved):
Confirmed the RFC index marked `RFC1384` as obsoleted by `RFC1617`:
```xml
  <rfc-entry>
    <doc-id>RFC1384</doc-id>
    <title>Naming Guidelines for Directory Pilots</title>
    ...
    <obsoleted-by>
      <doc-id>RFC1617</doc-id>
    </obsoleted-by>
    ...
  </rfc-entry>
```

tidoust:
Thanks for the review, @deniak.

@tobie, I'll merge this PR in the next couple of days unless I hear push back by then.

PR #936, opened 2026-06-23
COMPRESSION-DICTIONARY, DIGEST-HEADERS, and WEBP are now published RFCs

deniak:
The auto-update script has been failing because these 3 documents are now RFCs and are being dropped from `refs/browser-specs.json`


deniak:
@tobie thanks!

PR #937, opened 2026-06-23
fixes #932 by linking to the maintained svg2 specs

### LINKED ISSUE #932: svg spec links need to be updated.
https://github.com/search?q=repo%3Atobie%2Fspecref%20svgwg.org&type=code

Currently there are a couple of references to `svgwg.org` which makes in return bikeshed adds the wrong references in specs. 

The specs are now located at
https://w3c.github.io/svgwg/

karlcow:
the svgwg.org website is not maintained anymore. 
The fresh versions of the specs are on https://w3c.github.io/svgwg/

### DIFF
--- a/refs/browser-specs.json
+++ b/refs/browser-specs.json
@@ -1948,7 +1948,7 @@
         "repository": "https://github.com/WICG/storage-buckets"
     },
     "SVG-ANIMATIONS": {
-        "href": "https://svgwg.org/specs/animations/",
+        "href": "https://w3c.github.io/svgwg/specs/animations/",
         "title": "SVG Animations Level 2",
         "status": "Editor's Draft",
         "publisher": "W3C",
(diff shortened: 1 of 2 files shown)


tidoust (changes requested):
Sorry, I failed to comment on the underlying issue but:

1. That's not the right place to fix this, the W3C API would be the right source to fix. The scripts will overwrite the changes that this PR makes (except the change in github.js, but that's just logic to handle such URLs, it does not create data that targets svgwg.org)
2. The reason why these links are not fixed in the W3C system is that, so far, the SVG WG has not touched these other SVG specs (and it's not clear it would given the current charter). I can update the URL of SVG Animations in browser-specs.

karlcow:
@tidoust Thanks for the heads up! ❤️ 
What would happen if the domain svgwg.org was to disappear? Just dead links? I understand that the specs are not being updated, but the links are creating confusion. I still see that on a regular basis. 

What would be the best way to handle that? In a way that bikeshed would pick up the right references when generating spec references.

PR #942, opened 2026-07-15
add CTAP 2.3 reference

mohamedamir:
This PR adds the newly published FIDO CTAP 2.3 Proposed Standard published on February 26, 2026.


mohamedamir:
This should be imported automatically

PR #952, opened 2026-09-02
W3C renamed shacl12-rules to sparql12-rl

marcoscaceres:
W3C renamed shacl12-rules to sparql12-rl, so the hourly update now fetches the new shortname while the old entry stays behind and both claim the same 39 dated TR URLs, which is what test/dups.js has been failing on since 25 August. This renames the entry and leaves the old shortname and its dated versions as aliases, the same shape as b303ded5 for dx-prof.

Checked by running scripts/w3c.js against the W3C API both ways: 39 duplicate-URL failures before, none after, and it picks up the two versions published since the freeze. Written with Claude.

### DIFF
--- a/refs/w3c.json
+++ b/refs/w3c.json
@@ -183816,688 +183816,125 @@
         "repository": "https://github.com/w3c/data-shapes"
     },
     "shacl12-rules": {
-        "rawDate": "2026-08-19",
-        "href": "https://www.w3.org/TR/shacl12-rules/",
-        "title": "SHACL 1.2 Rules",
-        "edDraft": "https://w3c.github.io/data-shapes/shacl12-rules/",
-        "publisher": "W3C",
-        "source": "https://api.w3.org/specifications/shacl12-rules",
-        "versions": {
-            "20251209": {
-                "authors": [
-                    "Andy Seaborne",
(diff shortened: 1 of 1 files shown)


refs/w3c.json
```diff
@@ -183816,2534 +183816,2687 @@
         "repository": "https://github.com/w3c/data-shapes"
     },
     "shacl12-rules": {
-        "rawDate": "2026-08-19",
-        "href": "https://www.w3.org/TR/shacl12-rules/",
-        "title": "SHACL 1.2 Rules",
-        "edDraft": "https://w3c.github.io/data-shapes/shacl12-rules/",
+        "aliasOf": "sparql12-rl",
+        "isSeriesAlias": true
```
deniak, line 183820: This is not a series alias.
```suggestion
        "aliasOf": "sparql12-rl"
```

Instead, the series alias and renamed should be specify at line 182714:
```json
"shacl-rules": {
    "aliasOf": "sparql-rl",
    "isSeriesAlias": true
}
```

deniak (changes requested):
The diff showed in GitHub is messy because of all the updates. Running a local diff provides a much cleaner view.

